PR #49, opened 2014-11-21
Add renderClientOnly to Html.React to bypass server rendering

schotime:
Add this so that you can just render the placeholder for the react component where you want it and then still initialize the data to the component. 

Implementation was a bit hacky, so if you think there is a better way let me know.

You probably should also add a .gitattributes to specify you git settings (line endings etc)


Daniel15:
Closing this in favour of #103 (which has less copy-pasted code). Please feel free to comment on that pull request if it doesn't suit your needs and we can revisit this one. Thanks!

PR #54, opened 2014-11-28
Disable the source map generation if MsieJsEngine.

### LINKED ISSUE #53: Source map request return 500 when using MsieJavaScriptEngine
After long time investigation, the following statement returns different output in IE10 and IE11:

```
JSXTransformer.transform("var a = <span className=\"fuck\"/>", {sourceMap:true})
```

In IE11, which looks fine (notices that `sourceMap` object is generated):

```
{code: "var a = Rea...", extra: Object {...}, sourceMap: Object {...}, sourceMapFilename: "source.js"}
```

In IE10, suck... the `sourceMap` gone :(

```
{code: "var a = Rea...", extra: Object {...}}
```

When using MSIE JavaScript switcher in Json.Net, it is running in `ChakraActiveScript` mode, [its document][1] says `ChakraJsRt` mode leverage IE11. While I try with [this way][0], but it does not work, still in `ChakraActiveScript` mode.

Any idea on this?

[0]: https://github.com/Taritsyn/JavaScriptEngineSwitcher/blob/6a2dbb6f58e4fe2b7a4dc4b09dd204e001453c23/JavaScriptEngineSwitcher.Tests/App.config#L24

[1]: https://github.com/Taritsyn/MsieJavaScriptEngine/blob/master/README.md

---

Besides, I think when send transformed JSX, it should check if it can generate source map before set the JSX header.

lijunle:
Temporarily resolve #53 .

### DIFF
--- a/src/React.Web/JsxHandler.cs
+++ b/src/React.Web/JsxHandler.cs
@@ -66,9 +66,13 @@ namespace React.Web
 			var sourceMapUri = GetSourceMapUri(relativePath, result.Hash);
 			ConfigureCaching();
 			_response.ContentType = "text/javascript";
-			// The sourcemap spec says to use SourceMap, but Firefox only accepts X-SourceMap
-			_response.AddHeader("SourceMap", sourceMapUri);
-			_response.AddHeader("X-SourceMap", sourceMapUri);
+
+			if (this._environment.EngineSupportsSourceMapGeneration)
+			{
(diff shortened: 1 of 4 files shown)


lijunle:
Hi, @Daniel15 does any update for this PR?

Daniel15:
I think this is less of an issue now that ClearScript is included by default. I'm considering totally phasing out the MSIE engine in the future. What do you think?

lijunle:
OK, add V8 engine explicitly is good way.

PR #68, opened 2015-01-11
JsxTransformer exception now includes whole inner exception, not only message

mandrek44:
JsxTransformer adds the file name to exception message, however it discards the inner exception details, like stack trace. By including the inner exception in the thrown exception object we don't loose any information.

PR #70, opened 2015-01-12
JsxTransformer exception now includes whole inner exception, not only message

mandrek44:
JsxTransformer adds the file name to exception message, however it discards the inner exception details, like stack trace. By including the inner exception in the thrown exception object we don't loose any information.

Created again because original PR (https://github.com/reactjs/React.NET/pull/68) was made using wrong base branch.


Daniel15:
Thank you!

PR #73, opened 2015-01-25
Getting started corrections

rickbeerendonk:
Result of rebuilding the tutorial app: https://github.com/rickbeerendonk/ReactJS.NET-getting-started-tutorial/tree/1.3

### DIFF
--- a/site/jekyll/getting-started/tutorial.md
+++ b/site/jekyll/getting-started/tutorial.md
@@ -964,8 +964,6 @@ namespace ReactDemo
 	{
 		public static void Configure()
 		{
-			ReactSiteConfiguration.Configuration = new ReactSiteConfiguration();
-
 			ReactSiteConfiguration.Configuration
 				.AddScript("~/Scripts/showdown.js")
 				.AddScript("~/Scripts/Tutorial.jsx");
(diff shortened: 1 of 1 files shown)


Daniel15:
Thanks for doing that Rick! I've been meaning to do another run-through of the tutorial myself.

rickbeerendonk:
@Daniel15 Only adding bundling is a bit tough, because the BundleConfig.cs and its registration in Global.asax is not part of an Empty MVC project. The tutorial assumes it is though. You can see here what was needed to make it work: https://github.com/rickbeerendonk/ReactJS.NET-getting-started-tutorial/commit/540fd669dda4e6428481085495c5337c262c64ca

Daniel15:
The "ReactJS.NET - JSX for ASP.NET Web Optimization Framework" package should install ASP.NET Bundling and Minification, which should also create the BundleConfig.cs. Does it not do that?

I'm considering splitting that out into a separate page of the tutorial as soon I'm going to write some documentation about using Webpack instead.

rickbeerendonk:
I just tested it again in both a new MVC 4 and a new MVC 5 project, as well as in my tutorial code (MVC 4) and the BundleConfig.cs is not created. It creates a ReactConfig.cs, did you mean that one?

PR #106, opened 2015-03-18
Update download.md

rosdi:
Google's V8 engine is the default engine now, no need to install it anymore.


Daniel15:
You can actually completely remove any mention of that package, the entire package doesn't exist any more. The VroomJs one can be removed too.

rosdi:
Ok I have removed all references to the obsolete packages.

Daniel15:
Thanks!

PR #112, opened 2015-04-06
Disable caching of Comments() action

### LINKED ISSUE #111: Tutorial change to fix IE caching of comment ajax response
IE will aggressively cache ajax responses. In the tutorial the list of comments is retrieved from the server using the following method.

```csharp
public ActionResult Comments()
{
	return Json(_comments, JsonRequestBehavior.AllowGet);
}
```
This will not work out of the box in IE because the response is cached and will not contain the newly added comment.

To prevent IE from caching the response, you can add a Cache-Control header to the response like this:

```csharp
public ActionResult Comments()
{
	Response.Cache.SetCacheability(HttpCacheability.NoCache);

	return Json(_comments, JsonRequestBehavior.AllowGet);
}
```

andersea:
Adds an OutputCache attribute the the Comments() action, to prevent IE
from aggressively caching the comments list. Otherwise IE will only show
the comments that were returned in the first request unless you refresh
the entire page.

Fixes #111

### DIFF
--- a/site/jekyll/getting-started/tutorial.md
+++ b/site/jekyll/getting-started/tutorial.md
@@ -409,9 +409,9 @@ public ActionResult Comments()
 }
 ```
 
-The `OutputCache` attribute is used here to prevent IE from caching the ajax request. IE tries to optimize things by assuming that identical requests will return identical responses. Subsequent calls to the Comments action will simply return the cached response from the first call, the result being that the comment list is never updated in IE as new comments are added. When designing a real world API, cache times of API requests should be considered more carefully. For this tutorial it is easiest to simply disable caching.
+The OutputCache attribute is used here to prevent browsers from caching the response. When designing a real world API, caching of API requests should be considered more carefully. For this tutorial it is easiest to simply disable caching.
 
-Finally we add a corresponding route in `App_Start\RouteConfig.cs` to simplify the request URL a bit:
+Finally we add a corresponding route in `App_Start\RouteConfig.cs`:
 
 ```csharp{12-16}
 using System.Web.Mvc;
--- a/site/jekyll/getting-started/tutorial.md
+++ b/site/jekyll/getting-started/tutorial.md
@@ -402,13 +402,16 @@ namespace ReactDemo.Controllers
 Let's also add a new controller action to return the list of comments:
 
 ```csharp
+[OutputCache(Location = OutputCacheLocation.None)]
 public ActionResult Comments()
 {
 	return Json(_comments, JsonRequestBehavior.AllowGet);
 }
 ```
 
-And a corresponding route in `App_Start\RouteConfig.cs`:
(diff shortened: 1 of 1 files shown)


site/jekyll/getting-started/tutorial.md
```diff
@@ -402,13 +402,16 @@ namespace ReactDemo.Controllers
...
 ```csharp
+[OutputCache(Location = OutputCacheLocation.None)]
 public ActionResult Comments()
 {
 	return Json(_comments, JsonRequestBehavior.AllowGet);
 }
 ```
 
-And a corresponding route in `App_Start\RouteConfig.cs`:
+The `OutputCache` attribute is used here to prevent IE from caching the ajax request. IE tries to optimize things by assuming that identical requests will return identical responses. Subsequent calls to the Comments action will simply return the cached response from the first call, the result being that the comment list is never updated in IE as new comments are added. When designing a real world API, cache times of API requests should be considered more carefully. For this tutorial it is easiest to simply disable caching.
```
Daniel15: I think this description is a bit longer than it needs to be. I'd suggest changing this paragraph to:

> The `OutputCache` attribute is used here to prevent browsers from caching the response. When designing a real world API, caching of API requests should be considered more carefully. For this tutorial it is easiest to simply disable caching.

site/jekyll/getting-started/tutorial.md
```diff
@@ -402,13 +402,16 @@ namespace ReactDemo.Controllers
...
+
+Finally we add a corresponding route in `App_Start\RouteConfig.cs` to simplify the request URL a bit:
```
Daniel15: Remove the " to simplify the request URL a bit" bit - Ideally a production app shouldn't really use the default route.

Daniel15:
Thank you! Just some minor comments :) :+1:

site/jekyll/getting-started/tutorial.md
andersea: Why shouldn't a production app use the default route? I have done a couple of MVC apps and they both use default routes for almost everything. It seems to work fine.

I agree that the default route in this case looks a bit silly, but normally I think comments would be in another controller than the home controller, or maybe even a WebAPI call and that would fix the route automatically.

Daniel15: > Why shouldn't a production app use the default route?

It does work fine and you can do it that way, but ideally you should think about a good URL structure for pages in the app. Using the default route exposes internal implementation details (controller and action names) that you might want to change in the future. [Cool URIs don't change](http://www.w3.org/Provider/Style/URI.html), and well-designed URIs should remain the same regardless of the code behind them or the backend technology being used. Removing the default route makes you think about every URI as you build the site and generally results in cleaner more maintainable URIs.

I think at least some of that is personal preference though :P

andersea:
Thanks. Your suggestions have been applied.

Daniel15:
Thank you!

PR #121, opened 2015-05-04
Deferred JS Init

jhemminger:
Added functionality to allow React components to be rendered on the server side with the initialization data inline but stored to a global variable so that React is not immediately invoked.  At the end of the page the deferred init JavaScript function needs to be called which will initialize all components on the page.  This functionality is necessary for us to create components across different servers to comprise a page and allow us to load the React.JS script file at the end of the body and not in head.

### DIFF
--- a/src/React.AspNet/HtmlHelperExtensions.cs
+++ b/src/React.AspNet/HtmlHelperExtensions.cs
@@ -121,6 +121,38 @@ namespace React.AspNet
 			return new HtmlString(html + System.Environment.NewLine + script.ToString());
 		}
 
+        /// <summary>
+        /// Renders the specified React component, along with a script block used for
+        /// deferred initialization by <see cref="ReactInitDeferredJavaScript"/>
+        /// </summary>
+        /// <param name="htmlHelper">HTML helper</param>
+        /// <param name="componentName">Name of the component</param>
+        /// <param name="props">Props to initialise the component with</param>
+        /// <param name="htmlTag">HTML tag to wrap the component in. Defaults to &lt;div&gt;</param>
+        /// <param name="containerId">ID to use for the container HTML tag. Defaults to an auto-generated ID</param>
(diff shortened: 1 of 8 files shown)


PeteDuncanson:
Ouch my head...I have no idea what this is doing! :)

What do you mean by "create components across different servers to comprise a page"? It sounds like you are rendering different components on different servers but I can't see how that would be speedy or needed. Can you give me some more info?

jhemminger:
Sure I can explain more.  This library works great when you have a single MVC project, but for us this is not the case.  We have designed our site to have many separate functionally grouped units for which they each have their own release cadence/continuous delivery pipeline.  One challenge to this is the areas of the website that are standard across these separate pipelines, for instance the header and footer.  We want those components (which of course, are React objects) to be standardized across the site and be able to have their own continuous delivery pipeline so that we can updated them at will without having to deploy the functional pipelines that are using them in a coordinated fashion.  
So yes we are rendering different components on different servers, is it speedy as you asked?  Probably not as much  as if they were embedded with the same MVC project, but it's speediness, or lack there of is worth the benefits we get from this type of design.  I hope that makes a little more sense to you now, if not let me know!

jhemminger:
Crickets....Anyone else have any questions?  This is very similar to what the framework is already doing as a blend of ReactWithInit and ReactInitJavaScript, but it doesn't have the constraint of all calls needing to be in one web request context.

PeteDuncanson:
Hi. 

I'm happy with your description sorry for not acknowledging it. Love the idea of rendering on multiple servers. It sounds crazy but I like it :)

Pete

Daniel15:
Sorry I haven't had a chance to look at this yet. I'll try to take a look soon.

jhemminger:
OK thanks guys, I understand being busy and not having time I just wasn't sure what silence meant :)

PR #124, opened 2015-05-08
Bypass JSXTransforms load when not needed.

jhemminger:
From the recent addition of AddScriptWithoutTransform (which is great BTW, thanks!) it is likely, as is in my case, where I will not be loading any JSX scripts that require transformations.  So if the configuration contains no JSX scripts the JsxTransform script doesn't need to be loaded into the engine.  I may be missing something that this is needed for so please correct me if I am wrong.

Also I am not sure why the spacing looks off on these difs, it's correct in VS.

Thanks!
Joel

### DIFF
--- a/src/React.MSBuild/TransformJsx.cs
+++ b/src/React.MSBuild/TransformJsx.cs
@@ -52,8 +52,7 @@ namespace React.MSBuild
 			config
 				.SetReuseJavaScriptEngines(false)
 				.SetUseHarmony(UseHarmony)
-				.SetStripTypes(StripTypes)
-				.AddScript(string.Empty);
+				.SetStripTypes(StripTypes);
 
 			_environment = React.AssemblyRegistration.Container.Resolve<IReactEnvironment>();
 
--- a/src/React.MSBuild/TransformJsx.cs
+++ b/src/React.MSBuild/TransformJsx.cs
@@ -49,11 +49,11 @@ namespace React.MSBuild
 		{
 			MSBuildHost.EnsureInitialized();
 			var config = React.AssemblyRegistration.Container.Resolve<IReactSiteConfiguration>();
-		    config
-		        .SetReuseJavaScriptEngines(false)
-		        .SetUseHarmony(UseHarmony)
-		        .SetStripTypes(StripTypes)
-		        .AddScript(string.Empty);
+			config
+				.SetReuseJavaScriptEngines(false)
+				.SetUseHarmony(UseHarmony)
+				.SetStripTypes(StripTypes)
+				.AddScript(string.Empty);
 
 			_environment = React.AssemblyRegistration.Container.Resolve<IReactEnvironment>();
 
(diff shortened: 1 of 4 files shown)


Daniel15:
The spacing looks wrong because the line is now using a mix of tabs and spaces. Please change it so it's all tabs.

It looks like there was a build error, can you please run the full build + unit tests (`dev-build.bat`) and ensure it works?

src/React.MSBuild/TransformJsx.cs
```diff
@@ -49,10 +49,11 @@ public override bool Execute()
...
 			var config = React.AssemblyRegistration.Container.Resolve<IReactSiteConfiguration>();
-			config
-				.SetReuseJavaScriptEngines(false)
-				.SetUseHarmony(UseHarmony)
-				.SetStripTypes(StripTypes);
+		    config
+		        .SetReuseJavaScriptEngines(false)
+		        .SetUseHarmony(UseHarmony)
+		        .SetStripTypes(StripTypes)
+		        .AddScript(string.Empty);
```
Daniel15: Why is `string.Empty` necessary here? How does the code loading the scripts handle this?

jhemminger:
Fixed the spaces/tabs.  In regards to the build I will continue to investigate, running a rebuild all in VS works just fine, but the batch file blows up with something from Cassette.

src/React.MSBuild/TransformJsx.cs
```diff
@@ -52,7 +52,8 @@ public override bool Execute()
 			config
 				.SetReuseJavaScriptEngines(false)
 				.SetUseHarmony(UseHarmony)
-				.SetStripTypes(StripTypes);
+				.SetStripTypes(StripTypes)
+				.AddScript(string.Empty);
```
jhemminger: I had to add an item here because without it the MSBuild step would blow up.  I am not too familiar with what this build step is doing, but by making the loading of the JSX Transform conditional on having at least one Script caused this to blow up unless this was added.  If you can think of a more elegant solution that would be great!

jhemminger:
Well the build issue with Cassette led me down a much longer path than I hoped.  In the end I realized that ReactEnvironment has an instance of IJsxTransformer which can be invoked at will and expects the Jsx Transforms to be loaded into the engine.  So what I hoped to be an easy auto-magical no load of the transforms if not necessary isn't really possible.  I changed the load to be configured by a configuration parameter which should hopefully be obvious to someone that if they turn that off then try to use JsxTransformer it's not going to work.  If you don't want another configuration parameter I can understand, but it would be nice for those of us who don't need the transforms to not have them loaded into the engine.

Daniel15:
Perhaps you could load the transforms the first time they're used, if they
haven't already been loaded?

jhemminger:
I considered and attempted that for a bit, but I didn't see a clean way to do it.  It seemed I was cutting across too many different objects to make that happen and was very ugly.  I may just not know the code well enough to do it cleanly.

Daniel15:
I'm going to task this over and update it to work with Babel instead (#166). I think we can just throw an exception if the config option is disabled and the user tries to use `JsxTransformer`. Thanks for the pull request though!

Issue #127, opened 2015-05-19
Add an option to support renderToStaticMarkup method


xabikos:
Reactjs provides the method [renderToStaticMarkup][static] which is very useful when we need to render a page server side but not edit it client side. This method doesn't add all the extra react specific data-attributes so the produced markup is lot cleaner.

[static]: https://facebook.github.io/react/docs/top-level-api.html#react.rendertostaticmarkup

xabikos:
I am working on a pull request to resolve this

Daniel15:
Merged the pull request, thank you! :)

PR #128, opened 2015-05-20
Render react attributes

xabikos:
Added an optional parameter to the `Render` method of `IReactCompoment` interface with default value to true. This won't change the behavior of the existing applications. On the other hand gives the flexibility to not render the React data attributes during server side rendering if there is a need. Internally has implemented by calling `React.renderToStaticMarkup` method instead the default` React.renderToString`. I changed the corresponding method in `HtmlHelperExtensions` class. At last I didn't figure out how to make sure that my code lints. Is there a tool or something like this we have to execute? #127

### DIFF
--- a/site/jekyll/guides/server-side-rendering.md
+++ b/site/jekyll/guides/server-side-rendering.md
@@ -85,7 +85,7 @@ If thre is no need to have a React application client side and you just want to 
 })
 ```
 
-And the Html markup will look like which is a lot cleaner. In this case there is no need to laod the React script or call the `Html.ReactInitJavaScript()` method.
+And the Html markup will look like the one following which is a lot cleaner. In this case there is no need to laod the React script or call the `Html.ReactInitJavaScript()` method.
 
 ```html
 <div id="react1">
--- a/site/jekyll/guides/server-side-rendering.md
+++ b/site/jekyll/guides/server-side-rendering.md
@@ -72,3 +72,26 @@ meaning your initial render will be super fast.
 
 For a more in-depth example, take a look at the included sample application
 (**React.Samples.Mvc4**).
+
+5 - How not to render the React data attributes
+
+If thre is no need to have a React application client side and you just want to use the server side rendering but without the React specific data attributes call `Html.React` and pass renderReactAttributes parameter as false.
+
+```csharp
+@Html.React("HelloWorld", new
+{
+	name = "Daniel",
+ 	renderReactAttributes: false
+})
+```
+
+And the Html markup will look like which is a lot cleaner. In this case there is no need to laod the React script or call the `Html.ReactInitJavaScript()` method.
+
+```html
+<div id="react1">
+	<div>
+		<span>Hello </span>
+		<span>Daniel</span>
+	</div>
+</div>
+```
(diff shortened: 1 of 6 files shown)


Daniel15:
There's already a `clientOnly` setting that does not do server-side rendering, only client-side rendering. Perhaps this flag should follow a similar naming convention and be called `serverOnly`?

src/React.AspNet/HtmlHelperExtensions.cs
```diff
@@ -63,10 +63,11 @@ private static IReactEnvironment Environment
 		/// <typeparam name="T">Type of the props</typeparam>
 		/// <param name="htmlHelper">HTML helper</param>
 		/// <param name="componentName">Name of the component</param>
-		/// <param name="props">Props to initialise the component with</param>
```
Daniel15: "initialise" is correct in British English :) If you want to Americanise (Americanize?) the spelling, please do it in a separate pull request :)

site/jekyll/guides/server-side-rendering.md
```diff
@@ -72,3 +72,26 @@ meaning your initial render will be super fast.
...
+
+5 - How not to render the React data attributes
+
+If thre is no need to have a React application client side and you just want to use the server side rendering but without the React specific data attributes call `Html.React` and pass renderReactAttributes parameter as false.
+
+```csharp
+@Html.React("HelloWorld", new
+{
+	name = "Daniel",
+ 	renderReactAttributes: false
```
Daniel15: I don't think this example works as-is, you're defining `renderReactAttributes` inside the props rather than as a separate parameter.

src/React.AspNet/HtmlHelperExtensions.cs
```diff
@@ -63,10 +63,11 @@ private static IReactEnvironment Environment
...
+		/// <param name="props">Props to initialize the component with</param>
 		/// <param name="htmlTag">HTML tag to wrap the component in. Defaults to &lt;div&gt;</param>
 		/// <param name="containerId">ID to use for the container HTML tag. Defaults to an auto-generated ID</param>
-                /// <param name="clientOnly">Skip rendering server-side and only output client-side initialisation code. Defaults to <c>false</c></param>
+        /// <param name="clientOnly">Skip rendering server-side and only output client-side initialization code. Defaults to <c>false</c></param>
+        /// <param name="renderReactAttributes">Indicates if the React data-attributes should be rendered during server side rendering</param>
```
Daniel15: Please use tabs rather than spaces.

src/React.AspNet/HtmlHelperExtensions.cs
```diff
@@ -74,7 +75,8 @@ private static IReactEnvironment Environment
 			T props,
 			string htmlTag = null,
 			string containerId = null, 
-			bool clientOnly = false
+			bool clientOnly = false,
+            bool renderReactAttributes = true
```
Daniel15: Tabs

Daniel15:
> At last I didn't figure out how to make sure that my code lints.

I don't have any experience with C# linting tools. I'll probably try to set up FxCop or something similar at some point. One thing I did notice is that this diff is using spaces for indentation - Please use tabs instead :)

xabikos:
Thanks for the comments @Daniel15 I hope I fixed all the issues. If I sould change anything else please let me know.

Daniel15:
Thank you! Sorry I took a while to get to this, I was on a flight from the USA to Australia last week and haven't gotten around to checking emails and stuff until now.

xabikos:
Never micnd Daniel! Happy I helped!

PR #130, opened 2015-05-30
Updated to ASP.NET MVC 6 beta-4

st1pps:


src/React.AspNet/JsxFileSystem.cs
```diff
@@ -197,12 +197,12 @@ public DateTimeOffset LastModified
 
 			public void WriteContent(byte[] content)
 			{
-				_fileInfo.WriteContent(content);
+                File.WriteAllBytes(_fileInfo.PhysicalPath, content);
```
Daniel15: Please use tabs rather than spaces.

Also why does this need to be changed? What's wrong with the old version? Passing through to the inner `IFileInfo` should be sufficient.

src/React.Sample.Mvc6/React.Sample.Mvc6.xproj
```diff
@@ -1,10 +1,9 @@
-<?xml version="1.0" encoding="utf-8"?>
+﻿<?xml version="1.0" encoding="utf-8"?>
```
Daniel15: It looks like a space was added to the start here? Seems unnecessary.

src/React.Sample.Mvc6/Startup.cs
```diff
@@ -59,9 +59,7 @@ public void Configure(IApplicationBuilder app, IHostingEnvironment env, ILoggerF
 				// send the request to the following path or controller action.
 				app.UseErrorHandler("/Home/Error");
 			}
-
-			app.UseRequestServices();
```
Daniel15: If this is no longer required, can you please update the documentation too? https://github.com/reactjs/React.NET/blob/master/site/jekyll/getting-started/aspnet5.md

Daniel15:
Thanks for this! I was going to attempt it myself but NuGet is totally broken for me after upgrading from Visual Studio 2015 CTP to RC (https://github.com/NuGet/Home/issues/692), and I haven't had time to try a total reinstall

Daniel15:
Finally got NuGet working again. Merged this with some minor changes (removed the  modified  methods from `JsxFileSystem` as they're not actually needed. Thanks!

Daniel15:
Released 1.5.2 with this update.

st1pps:
Sorry for the unecessary changes, and thank you very much for reviewing and merging.

Issue #122, opened 2015-05-05
Server-side `console.log`s before client-side


SoonDead:
I love the current console shim. It is great to be able to output messages on the server.

What could be improved is the ordering of server and client-side messages. The server-side messages should show up on the console before client-side rendering happens.

Might be solvable with a simple reordering of `GetInitJavaScript()`, where the `consoleCalls` are appended before the components.

Daniel15:
Yeah, you should simply be able to swap the order such that the console calls are first: https://github.com/reactjs/React.NET/blob/master/src/React.Core/ReactEnvironment.cs#L293-301

Feel free to send a pull request :)

Daniel15:
This was merged in #133

PR #133, opened 2015-06-02
Switch the order of server and client-side console.log calls

SoonDead:
Server-side calls should show up before their client-side counterparts, as server-side rendering runs first.


SoonDead:
This is supposed to fix #122

Daniel15:
Thanks! Please fill in the CLA and I can merge this :)

SoonDead:
Okay.

PR #157, opened 2015-08-21
Fix name of build batch file & sample avatars

radnor:
Document still referred to the old filename.

### DIFF
--- a/site/jekyll/getting-started/download.md
+++ b/site/jekyll/getting-started/download.md
@@ -47,7 +47,7 @@ Building Manually
 To build your own copy of ReactJS.NET (for example, if implementing a new
 feature or fixing a bug):
 
-1. Compile ReactJS.NET by running `build.bat`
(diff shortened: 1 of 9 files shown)


radnor:
Added in a fix to the samples so they use github avatar pictures.

Daniel15:
Thank you! For the avatars, Github also have URLs with the username (eg. https://avatars.githubusercontent.com/Daniel15?s=50), which looks nicer than having an ID in the URL :smile:  

Do you know if this is documented on the Github site anywhere? Is it possible it'd break in the future?

radnor:
Ah, nice, that URL is friendlier. :smile: 

I can't find it documented anywhere, but if it does disappear in the future it'll be easy enough to re-implement using the ID. I'll submit another commit shortly to utilize usernames.

Daniel15:
Thanks! Sorry there's duplicate code all over the place with the example; I haven't spent much time on them.

PR #165, opened 2015-10-09
adding server side paging without javascript

samwa:
Hi
I've fixed up the React.Sample.Mvc4 project so that with javascript turned off, the paging still works
previously if js was disabled it would just return the json, now the controller listens to the request content  type and will respond as json or returns the index view for the next page

### DIFF
--- a/src/React.Sample.Mvc4/Content/Sample.jsx
+++ b/src/React.Sample.Mvc4/Content/Sample.jsx
@@ -9,12 +9,13 @@
 
 var CommentsBox = React.createClass({
 	propTypes: {
-		initialComments: React.PropTypes.array.isRequired
+		initialComments: React.PropTypes.array.isRequired,
+		page: React.PropTypes.number
 	},
 	getInitialState() {
 		return {
 			comments: this.props.initialComments,
-			page: 1,
+			page: this.props.page,
 			hasMore: true,
 			loadingMore: false
 		};
@@ -27,8 +28,9 @@ var CommentsBox = React.createClass({
 		});
 
 		var url = evt.target.href;
-		var xhr = new XMLHttpRequest();
+		var xhr = new XMLHttpRequest();			
 		xhr.open('GET', url, true);
+		xhr.setRequestHeader('Content-Type', 'application-json');
 		xhr.onload = () => {
 			var data = JSON.parse(xhr.responseText);
 			this.setState({
@@ -60,7 +62,7 @@ var CommentsBox = React.createClass({
 			return <em>Loading...</em>;
 		} else if (this.state.hasMore) {
 			return (
-				<a href={'comments/page-' + (this.state.page + 1)} onClick={this.loadMoreClicked}>
+				<a href={'/comments/page-' + (this.state.page + 1)} onClick={this.loadMoreClicked}>
 					Load More
 				</a>
 			);
--- a/src/React.Sample.Mvc4/Content/Sample.jsx
+++ b/src/React.Sample.Mvc4/Content/Sample.jsx
@@ -28,9 +28,9 @@ var CommentsBox = React.createClass({
 		});
 
 		var url = evt.target.href;
-		var xhr = new XMLHttpRequest();			
+		var xhr = new XMLHttpRequest();
 		xhr.open('GET', url, true);
-		xhr.setRequestHeader('Content-Type', 'application-json');
+		xhr.setRequestHeader('Content-Type', 'application/json');
 		xhr.onload = () => {
 			var data = JSON.parse(xhr.responseText);
 			this.setState({
--- a/src/React.Sample.Mvc4/Controllers/HomeController.cs
+++ b/src/React.Sample.Mvc4/Controllers/HomeController.cs
@@ -36,6 +36,7 @@ namespace React.Sample.Mvc4.ViewModels
 	{
 		public IEnumerable<CommentModel> Comments { get; set; }
 		public int CommentsPerPage { get; set; }
+        public int Page { get; set; }
 	}
 }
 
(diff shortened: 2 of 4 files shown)


src/React.Sample.Mvc4/Content/Sample.jsx
```diff
@@ -27,8 +28,9 @@ var CommentsBox = React.createClass({
 		});
 
 		var url = evt.target.href;
-		var xhr = new XMLHttpRequest();
+		var xhr = new XMLHttpRequest();
```
Daniel15: Please remove the excess whitespace here

src/React.Sample.Mvc4/Content/Sample.jsx
```diff
@@ -27,8 +28,9 @@ var CommentsBox = React.createClass({
 		});
 
 		var url = evt.target.href;
-		var xhr = new XMLHttpRequest();
+		var xhr = new XMLHttpRequest();			
 		xhr.open('GET', url, true);
+		xhr.setRequestHeader('Content-Type', 'application-json');
```
Daniel15: Correct MIME type is `application/json` (`/` not `-`)

src/React.Sample.Mvc4/Controllers/HomeController.cs
```diff
@@ -36,6 +36,7 @@ public class IndexViewModel
 	{
 		public IEnumerable<CommentModel> Comments { get; set; }
 		public int CommentsPerPage { get; set; }
+        public int Page { get; set; }
```
Daniel15: Please fix indentation - You're using spaces, but this project uses tabs.

src/React.Sample.Mvc4/Controllers/HomeController.cs
```diff
@@ -77,7 +78,8 @@ public ActionResult Index()
             return View(new IndexViewModel
             {
 	            Comments = _comments.Take(COMMENTS_PER_PAGE),
-				CommentsPerPage = COMMENTS_PER_PAGE
+				CommentsPerPage = COMMENTS_PER_PAGE,
+                Page = 1
```
Daniel15: Please fix indentation - You're using spaces, but this project uses tabs.

src/React.Sample.Mvc4/Controllers/HomeController.cs
```diff
@@ -86,10 +88,25 @@ public ActionResult Comments(int page)
...
+	            {
+	                comments = comments,
+	                hasMore = hasMore
+	            }, JsonRequestBehavior.AllowGet);
+	        }
+	        else
+	        {
+	            return View("Index", new IndexViewModel
+	            {
+	                Comments = _comments.Take(COMMENTS_PER_PAGE * page),
```
Daniel15: Fix indentation

src/React.Sample.Mvc4/Controllers/HomeController.cs
```diff
@@ -86,10 +88,25 @@ public ActionResult Comments(int page)
 		    var comments = _comments.Skip((page - 1) * COMMENTS_PER_PAGE).Take(COMMENTS_PER_PAGE);
 			var hasMore = page * COMMENTS_PER_PAGE < _comments.Count;
 
-		    return Json(new {
-				comments = comments,
-				hasMore = hasMore
-			}, JsonRequestBehavior.AllowGet);
+	        if (ControllerContext.HttpContext.Request.ContentType == "application-json")
```
Daniel15: I would prefer if there were two separate actions - One that returns JSON, and one that returns HTML. The `loadMoreClicked` function in JS can hit the JSON endpoint, while the <a>'s href can be the HTML one.

Alternatively, switching by content-type is fine, but you **must** send a `Vary: Content-Type` header to prevent proxies from incorrectly caching it. I think the OutputCache attribute would let you do that.

samwa: thanks
i'm going to add the output cache option as having 2 endpoints feels like duplication to me. and the code that generates the loadMoreClicked uses the Url of the link, so the JSX would have to be 'fudged' to make it work
mvc 6 is going to have better content negotiation built in

samwa:
Appveyor seems to be getting the wrong version of VroomJs
I can't figure out what I might of changed that would have caused this error, sorry

src/React.Sample.Mvc4/Controllers/HomeController.cs
```diff
@@ -69,27 +71,43 @@ public HomeController()
...
+			});
+		}
 
-	    public ActionResult Comments(int page)
-	    {
-		    var comments = _comments.Skip((page - 1) * COMMENTS_PER_PAGE).Take(COMMENTS_PER_PAGE);
+		[OutputCache(Duration = 0, Location = OutputCacheLocation.Any, VaryByHeader = "Content-Type")]
+		public ActionResult Comments(int page)
+		{
+			Response.Cache.SetOmitVaryStar(true);
```
Daniel15: Why is this needed?

Daniel15:
Thanks, this looks pretty good. Can you please squash your commits?

src/React.Sample.Mvc4/Controllers/HomeController.cs
samwa: Do you man the line about omit vary by star? there is a pretty esoteric reason why iis won't send the cache header to the response when you use varybyheader. I don't understand it fully, but testing the examples given by this SO thread worked
http://stackoverflow.com/questions/7849392/unable-to-append-vary-header-to-response

samwa:
Hi Dan, sorry, I'm new to Git and contributing to OS projects. I had a little trouble rebase-ing my commits. long and the short of it is I ended up creating a branch feature-serversidepaging with one commit. I'll try a new pull request from that

Daniel15:
Thanks! I'll close this one and take a look at the other pull request.

PR #170, opened 2015-10-15
Fix broken tests in BabelTransformerTests.cs

radnor:
Change two `Assert.AreEqual()` calls to `StringAssert.EndsWith()`. This fixes the build.


Daniel15:
How is the test failing - What's in `result` when it fails?

radnor:
Here's what's in `result`:
```
// Automatically generated by ReactJS.NET. Do not edit, your changes will be overridden.
// Version: 
// Generated at: 10/15/2015 9:02:38 AM
///////////////////////////////////////////////////////////////////////////////
React.DOM.div('Hello World')
```

Daniel15:
Ahh, this must have broken with 4c73590b85f6d81e37e012d9b9eb5dd52bd1d148. derp. Thanks for fixing it.

PR #175, opened 2015-10-19
Change React.render() to ReactDOM.render()

radnor:
I updated `RenderJavaScript()` in `src/React.Core/ReactComponent.cs` to call `ReactDOM.render()` per the v0.14 update, and modified all instances in the markdown docs.

I also noticed the OWIN sample wasn't running properly so I fixed that too. :smile:

### DIFF
--- a/site/jekyll/getting-started/tutorial.md
+++ b/site/jekyll/getting-started/tutorial.md
@@ -100,7 +100,7 @@ var CommentBox = React.createClass({
     );
   }
 });
-React.render(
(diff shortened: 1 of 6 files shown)


Daniel15:
Thanks for updating this! For the tutorial, it might be worth diffing the ReactJS.NET version to the official React version (https://github.com/facebook/react/blob/master/docs/docs/tutorial.md) to see if there's any other updates I've missed.

Issue #182, opened 2015-11-03
Using MD5 for the FileCacheHash causes exception to be throw if FIPS compliance group policy is enabled


RPrimrose:
Hi,
we recently tried installing an application using ReactJS.NET on a machine with the "System cryptography: Use FIPS-compliant algorithms for encryption, hashing, and signing" Group Policy enabled. There's more information on that policy here https://support.microsoft.com/en-us/kb/811833

The policy is causing the following exception to be thrown "System.InvalidOperationException: This implementation is not part of the Windows Platform FIPS validated cryptographic algorithm" when trying to render components.

The issue is that in FileCacheHash.cs the hash algorithm used is MD5 and this isn't certified as FIPS compliant. 

I believe it's a one-line change to fix this, I've tested changing creation of _hash to
`private readonly HashAlgorithm _hash = SHA512.Create("System.Security.Cryptography.SHA512CryptoServiceProvider");` and that seems to work fine. 

If that change sounds reasonable let me know and I'll submit a pull request. 

(You can recreate this error by following the instructions to turn on the group policy at http://stackoverflow.com/a/13635742 and then trying to render any MVC page that includes a @HTML.React() line in it.)

Daniel15:
Interesting, I didn't know about that setting. I used MD5 because it's fast and fine to use for file hashing (doesn't need to be cryptographically secure) but it's probably fine to change it to SHA512 instead. Feel free to send a pull request :smile:

Do you know which algorithms are allowed when FIPS-compliant algorithms are enforced? I think using whichever one is fastest would be sufficient for this use case.

RPrimrose:
Yeah, it was a new one to me as well. 

Strangely MSDN doesn't seem to have an official list of FIPS-compliant algorithms but http://blogs.msdn.com/b/icumove/archive/2009/01/31/working-with-fips-in-net-c.aspx has some info.

I've gone with SHA1Cng in the pull-request, should be a bit faster than SHA512.

PeteDuncanson:
This ones gold dust. Great find.

Daniel15:
Fixed in 96b6f56eba05ea0ab52183fee460b5e7a7392f12, thanks @RPrimrose!

PR #184, opened 2015-11-04
Amend FileCacheHash for FIPS compliance by using SHA1 instead of MD5

RPrimrose:
Changing FileCacheHash to use a FIPS compliant algorithm for https://github.com/reactjs/React.NET/issues/182

### DIFF
--- a/src/React.Core/FileCacheHash.cs
+++ b/src/React.Core/FileCacheHash.cs
@@ -24,9 +24,9 @@ namespace React
 		private const string HASH_PREFIX = "// @hash v3-";
 
 		/// <summary>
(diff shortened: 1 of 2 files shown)


src/React.Core/FileCacheHash.cs
```diff
@@ -24,9 +24,9 @@ public class FileCacheHash : IFileCacheHash
 		private const string HASH_PREFIX = "// @hash v3-";
 
 		/// <summary>
-		/// Althorithm for calculating file hashes
+		/// Algorithm for calculating file hashes
```
Daniel15: LOL, thanks :laughing:

PR #211, opened 2016-01-16
ContainerClass

jovnas:
Add possibility to add class to container tag.

### DIFF
--- a/src/React.Core/ReactComponent.cs
+++ b/src/React.Core/ReactComponent.cs
@@ -53,6 +53,11 @@ namespace React
 		public string ContainerTag { get; set; }
 
 		/// <summary>
+		/// Gets or sets the HTML class for the container of this component
+		/// </summary>
+		public string ContainerClass { get; set; }
+
+		/// <summary>
 		/// Gets or sets the props for this component
 		/// </summary>
 		public object Props { get; set; }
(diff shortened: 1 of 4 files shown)


src/React.Core/ReactComponent.cs
```diff
@@ -94,12 +99,25 @@ public virtual string RenderHtml(bool renderContainerOnly = false, bool renderSe
...
+						"<{2} id=\"{0}\">{1}</{2}>",
+						ContainerId,
+						html,
+						ContainerTag
+						);
+				}
+				else
+				{
+					return string.Format(
+						"<{2} id=\"{0}\" class=\"{3}\">{1}</{2}>",
```
Daniel15: Rather than having two separate branches here, it might be worth having a list of attributes

Something like (not tested):
```csharp
var attributes = id=\"" + ContainerId + "\";
if (!string.IsNullOrEmpty(ContainerClass))) {
	attributes += " class=\"" + ContainerClass + "\"";
}

return string.Format(
	"<{2} {0}>{1}</{2}>",
	attributes,
	html,
	ContainerTag
);
```

Daniel15:
Thanks! This looks reasonable to me. Could you please squash your commits, and then I'll merge this :)

jovnas:
I think I got it right... Haven't really used git before. :-)

Daniel15:
Thanks!

Daniel15:
This is now available in [ReactJS.NET 2.2](http://reactjs.net/2016/01/2.2.0-release.html). Thanks for your contribution!

jovnas:
Nice! Thank you!

PR #217, opened 2016-01-25
Added brief guide for running ReactJS.NET in Azure.

samppis:
ReactJS.NET version 2.2.0 fixes all issues related to running the V8 JavaScript engine in Azure. I still wrote a brief documentation so that people will know it works and that a version update might be necessary. I also included code to check the available JavaScript engines and how to disable the MSIE engine in case someone is still having problems with V8.

Please feel free to modify the text or fix typos etc. if necessary!

### DIFF
--- a/site/jekyll/guides/azure.md
+++ b/site/jekyll/guides/azure.md
@@ -1,0 +1,43 @@
+---
+layout: docs
+title: Azure
+---
+
+From **version 2.2.0** onwards ReactJS.NET works out of the box in Azure using the V8 JavaScript engine. Versions prior to 2.2.0 will fall back to using MSIE JavaScript engine and you may experience JavaScript errors during server-side rendering that you aren't experiencing locally.
+
+You can run the following code to check which JavaScript engines are available on the machine that your application is running on. The engine with the lowest priority is used by ReactJS.NET.
+
+```csharp
+public string GetAvailableEngines()
+{
+    var sb = new StringBuilder();
+    var registrations = React.TinyIoC.TinyIoCContainer.Current.ResolveAll<JavaScriptEngineFactory.Registration>();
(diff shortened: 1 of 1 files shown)


site/jekyll/guides/azure.md
```diff
@@ -0,0 +1,43 @@
...
+    }
+
+    return sb.ToString();
+}
+```
+
+To force ReactJS.NET to use the V8 JavaScript engine (and throw an exception if it isn't available) set the AllowMsieEngine configuration property to false.
+
+```csharp
+app.UseReact(config =>
```
Daniel15: It's probably worth mentioning that this syntax (`UseReact`) is specific to ASP.NET 5, and users of ASP.NET 4 will need to do this in their `ReactConfig.cs` file intsead.

site/jekyll/guides/azure.md
```diff
@@ -0,0 +1,43 @@
+---
+layout: docs
+title: Azure
+---
+
+From **version 2.2.0** onwards ReactJS.NET works out of the box in Azure using the V8 JavaScript engine. Versions prior to 2.2.0 will fall back to using MSIE JavaScript engine and you may experience JavaScript errors during server-side rendering that you aren't experiencing locally.
```
Daniel15: Nit: Just say `2.2` rather than `2.2.0`, I usually omit the `.0` :)

Daniel15:
This looks pretty good. I just posted some small comments inline. Thank you so much! :smile:

Daniel15:
Thanks, I pushed this to the live site: http://reactjs.net/guides/azure.html

This makes me think that I should probably add a better way to get a list of all the available engines...

samppis:
Great, thanks! That might be a good idea ;)

2016-01-25 14:10 GMT+07:00 Daniel Lo Nigro <notifications@github.com>:

PR #224, opened 2016-02-11
Update Copyrights to Present

lazytype:
Might be better to be more future-proof by saying "2014-Present" instead.


Daniel15:
I like the idea of "2014-Present", would you be able to update this PR to use that instead? :smile:

lazytype:
Sure

lazytype:
Updated :)

Daniel15:
Manually merged in df0313d652b339b1f6cdbf59811da219611c0973 due to a conflict in one file. Thanks for your contribution, and sorry it took me so long to reply!

PR #226, opened 2016-02-15
Prevent ReactComponent to serialize Props twice on one http request.

sfadeev:
Small performance improvement to prevent ReactComponent to serialize `Props` twice on one http request:
* first time on `@Html.React` call
* second time on `@Html.ReactInitJavaScript` call

### DIFF
--- a/src/React.Core/ReactComponent.cs
+++ b/src/React.Core/ReactComponent.cs
@@ -38,6 +38,11 @@ namespace React
 		protected readonly IReactSiteConfiguration _configuration;
 
 		/// <summary>
+		/// Serialized <see cref="Props" />
+		/// </summary>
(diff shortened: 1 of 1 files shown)


Daniel15:
The issue with this approach is that `Props` is not immutable. In theory it could be changed between the `Html.React` and `Html.ReactInitJavaScript` calls. Even though that's pretty silly, someone may be doing it in their code. I think instead, `Props` should have a custom setter that also sets the serialized props.

Daniel15:
I used an alternate approach in cdaf7372a364f176f1dbc2f0b4c6f3eaf1a3d75f (serialize the props in the setter rather than in `GetComponentInitialiser`. Thanks for the pull request though, it was a good inspiration :)

PR #239, opened 2016-03-07
Adding ability to render script literals

abezydar:
Perhaps this is helpful for others, I needed to be able to add script string literals to be processed, rather than file names for a Class Library I'm working on as I'll have bundled resource files or will be grabbing the scripts asynchronously through an API.  

The relevant methods were added to **IReactSiteConfiguration** and its implementation **ReactSiteConfiguration**, with a simple test added to the Console App Sample. **Babel** was also updated to handle transforming and caching a script literal. I took the approach of using a Dictionary to store the script literal in the ReactSiteConfiguration where the key ends up as the cache key. 

A couple notes and thoughts:
I could probably combine more of the logic for caching in the Babel class as well as transforming without a sourcemap. I haven't tested **AddScriptLiteralWithoutTransform**.


Daniel15:
Could you please adjust the indentation? This project uses tabs for indentation, not spaces. All the lines you've modified are using spaces :smile:  If you install the [EditorConfig Visual Studio extension](http://editorconfig.org/), it will automatically use the correct settings.

It looks like some unit tests are failing: https://ci.appveyor.com/project/Daniel15/react-net/build/89

src/React.AspNet/project.json
```diff
@@ -1,5 +1,5 @@
 {
-    "version": "2.2.2-*",
+    "version": "2.2.2-dev-20160307-1719",
```
Daniel15: Please revert these changes. `dev-build.bat` should automatically revert them back after a successful build.

src/React.Core/Babel.cs
```diff
@@ -79,6 +79,53 @@ public Babel(IReactEnvironment environment, ICache cache, IFileSystem fileSystem
 			_babelConfig = siteConfig.BabelConfig.Serialize();
 		}
 
+        /// <summary>
+        /// Transforms a JavaScript literal. Results of the transformation are cached.
+        /// </summary>
+        /// <param name="scriptId">The id of the script, used for caching purposes</param>
+        /// <param name="script">The script contents</param>
+        /// <returns>JavaScript</returns>
+        public virtual string TransformLiteral(string scriptId, string script)
```
Daniel15: This name is pretty confusing, what about calling it `TransformWithSourceMapCached` or something like that?

Also, please add unit tests for this method.

src/React.Core/Babel.cs
```diff
@@ -79,6 +79,53 @@ public Babel(IReactEnvironment environment, ICache cache, IFileSystem fileSystem
...
+
+            // 3. Cache the result from above to memory
+            _cache.Set(
+                cacheKey,
+                output,
+                slidingExpiration: TimeSpan.FromMinutes(30)                
+            );
+
+            return output.Code;
+
```
Daniel15: Remove the blank lines

src/React.Core/Babel.cs
```diff
@@ -147,6 +193,23 @@ public virtual string TransformFile(string filename)
...
+        /// Checks the in-memory cache. We need to invalidate any in-memory cache if there's no 
+        /// source map cached and forceGenerateSourceMap is true.
+        /// </summary>
+        /// <param name="cacheKey"></param>
+        /// <param name="forceGenerateSourceMap"></param>
+        /// <param name="cached"></param>
+        /// <returns>True if the cache contains the transformed source.</returns>
+        private bool CheckInMemoryCache(string cacheKey, bool forceGenerateSourceMap, out JavaScriptWithSourceMap cached)
+        {
+            cached = _cache.Get<JavaScriptWithSourceMap>(cacheKey);
```
Daniel15: Instead of having an out parameter, what about just returning `null` if the cache isn't valid?

```
var cached = _cache.Get<JavaScriptWithSourceMap>(cacheKey);
return cached != null && (!forceGenerateSourceMap || cached.SourceMap != null
  ? cached
  : null;
```

src/React.Core/IReactSiteConfiguration.cs
```diff
@@ -43,17 +43,48 @@ public interface IReactSiteConfiguration
...
+        /// <summary>
+        /// Adds a script literal to the list of scripts that are executed. This is the same as
+        /// <see cref="AddScriptLiteral"/> except it does not run JSX transformation on the script and thus is
+        /// more efficient.
+        /// </summary>
+        /// <param name="script">
+        /// The script literal to execute.
+        /// </param>
+        /// <returns>The configuration, for chaining</returns>
+        IReactSiteConfiguration AddScriptLiteralWithoutTransform(string script);
```
Daniel15: Do you think this is useful? It seems like overkill :) How about just adding `AddScriptLiteral` for now?

abezydar:
No problem Daniel, I'll make the requested changes!

src/React.Core/package.json
```diff
@@ -1,19 +1,20 @@
...
+	"version": "1.0.0",
+	"name": "React.Core",
+	"private": true,
+	"devDependencies": {
+		"babel-core": "^6.3.13",
+		"babel-loader": "^6.2.0",
+		"babel-preset-es2015": "^6.3.13",
+		"babel-preset-stage-0": "^6.3.13",
+		"babel-standalone": "^1.0.0",
+		"babel-plugin-transform-regenerator": "^6.3.0",
```
abezydar: I had to include this line in order to get the build to work properly. I can remove it if it's unnecessary.

Daniel15: Hmm, strange. What error do you get if you remove this? In theory it shouldn't be needed at all.

abezydar: Hrm, I removed it this morning and rebuilt with no errors. Earlier, I was getting an error saying it was required and the build wouldn't continue. I'll commit the change.

mwethington:
failing unit tests

dustinsoftware:

PR #247, opened 2016-03-15
Update tutorial.md

jgoliver:
Adding explanation of replacing componentWillMount() with loadCommentsFromServer() and how that is then used in componentDidMount().


Daniel15:
Thanks!

PR #254, opened 2016-04-15
Allow Javascript engines to be bypassed entirely.

dustinsoftware:
If no server-side rendering will occur on a page, for instance if
component render is called with clientOnly: true, it is
pointless to initialize a Javascript engine on the server. This allows the
consumer to easily disable all server side Javascript if the server is
under heavy load.

### DIFF
--- a/src/React.AspNet/HtmlHelperExtensions.cs
+++ b/src/React.AspNet/HtmlHelperExtensions.cs
@@ -81,7 +81,7 @@ namespace React.AspNet
 			string containerClass = null
 		)
 		{
-			var reactComponent = Environment.CreateComponent(componentName, props, containerId);
+			var reactComponent = Environment.CreateComponent(componentName, props, containerId, clientOnly);
 			if (!string.IsNullOrEmpty(htmlTag))
 			{
 				reactComponent.ContainerTag = htmlTag;
@@ -118,7 +118,7 @@ namespace React.AspNet
 			string containerClass = null
 		)
 		{
-			var reactComponent = Environment.CreateComponent(componentName, props, containerId);
+			var reactComponent = Environment.CreateComponent(componentName, props, containerId, clientOnly);
 			if (!string.IsNullOrEmpty(htmlTag))
 			{
 				reactComponent.ContainerTag = htmlTag;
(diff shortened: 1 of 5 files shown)


dustinsoftware:
I see that I forgot to sign the CLA before submitting this PR - just did that.

Daniel15:
Seems reasonable to me! 

It looks like there's a number of "invisible" changes that aren't actually related to this diff (looks like it might be changing the line endings from `\r\n` to `\n` or vice versa. Could you please revert those? We can fix those up in a separate commit :)

src/React.AspNet/HtmlHelperExtensions.cs
```diff
@@ -5,26 +5,26 @@
...
+using IHtmlHelper = System.Web.Mvc.HtmlHelper;
 #else
 using Microsoft.AspNet.Mvc.Rendering;
 using IHtmlString = Microsoft.AspNet.Html.Abstractions.IHtmlContent;
 #endif
-
+
 #if LEGACYASPNET
-namespace React.Web.Mvc
+namespace React.Web.Mvc
```
Daniel15: I can't see any noticeable change in this line so it's probably replaced newlines (`\r\n` with `\n` or vice versa). Could you please revert all the lines that aren't directly related to your change?

dustinsoftware:
Yeah, it looks like the line endings are mixed in some of the files and VS auto corrected them :) Fix pushed

Daniel15:
It looks like a unit test is failing.  Can you please take a look? https://ci.appveyor.com/project/Daniel15/react-net/build/108

> Mvc\HtmlHelperExtensionsTests.cs(39,27): error CS0854: An expression tree may not contain a call or invocation that uses optional arguments [C:\projects\react-net\src\React.Tests\React.Tests.csproj]
> Mvc\HtmlHelperExtensionsTests.cs(64,27): error CS0854: An expression tree may not contain a call or invocation that uses optional arguments [C:\projects\react-net\src\React.Tests\React.Tests.csproj]
> Mvc\HtmlHelperExtensionsTests.cs(86,27): error CS0854: An expression tree may not contain a call or invocation that uses optional arguments [C:\projects\react-net\src\React.Tests\React.Tests.csproj]
> Done Building Project "C:\projects\react-net\src\React.Tests\React.Tests.csproj" (Rebuild target(s)) -- FAILED.

You can run the `dev-build.bat` file which will build the project and run the unit tests. Please ensure the bundled example site (`React.Sample.Mvc4`) still works fine as well.

dustinsoftware:
Fixed, sorry about that. I had been using the sample site to test before, but forgot to build solution before I pushed.

Daniel15:
Thanks 😄

dustinsoftware:
@Daniel15 would you mind publishing a package update so I can consume these changes?

Daniel15:
Will try to do that soon :)  In the meantime, you could use the latest development package (see http://reactjs.net/getting-started/download.html#development-builds)

Daniel15:
This has been included in the 2.4 release (http://reactjs.net/2016/05/2.4.0-release.html). Thanks for your contribution!

PR #270, opened 2016-05-19
Always return JS engine to pool after component render.

dustinsoftware:
This allows other threads to use available engines while other MVC actions are being processed during a page render.

### DIFF
--- a/src/React.AspNet/HtmlHelperExtensions.cs
+++ b/src/React.AspNet/HtmlHelperExtensions.cs
@@ -81,17 +81,24 @@ namespace React.AspNet
 			string containerClass = null
 		)
 		{
-			var reactComponent = Environment.CreateComponent(componentName, props, containerId, clientOnly);
-			if (!string.IsNullOrEmpty(htmlTag))
+			try
 			{
-				reactComponent.ContainerTag = htmlTag;
+				var reactComponent = Environment.CreateComponent(componentName, props, containerId, clientOnly);
+				if (!string.IsNullOrEmpty(htmlTag))
+				{
+					reactComponent.ContainerTag = htmlTag;
+				}
+				if (!string.IsNullOrEmpty(containerClass))
+				{
+					reactComponent.ContainerClass = containerClass;
+				}
+				var result = reactComponent.RenderHtml(clientOnly, serverOnly);
+				return new HtmlString(result);
 			}
-			if (!string.IsNullOrEmpty(containerClass))
+			finally
 			{
-				reactComponent.ContainerClass = containerClass;
+				Environment.ReturnEngineToPool();
 			}
-			var result = reactComponent.RenderHtml(clientOnly, serverOnly);
-			return new HtmlString(result);
 		}
 
 		/// <summary>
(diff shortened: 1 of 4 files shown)


huan086:
On some of my pages, I render 2 or 3 separate components (i.e. 2 or 3 Html.React calls). Some of these calls are in Html.Partial. Will this change cause any issues?

dustinsoftware:
Nope, you should be just fine.

Daniel15:
Thanks! The only issue I can think of is if someone is relying on global state - For instance, if one component's `render` function sets a global variable, and then another component expects that variable to be set. I'd like to think that that's a rare use-case though. And in cases like that, the user can just disable engine pooling.

Could you please add a unit test, at least for the `ReturnEngineToPool` method (and ideally the HTML helpers, if it's not too difficult to write tests for them)?

dustinsoftware:
FWIW we've been running this fix in production for a week now and have had no problems so far! 🎉

mwethington:
dustin - how is it going after a month or so?

dustinsoftware:
Really well so far. One time one of the pooled v8 engines got in a weird
state and stopped responding - we had to recycle the app to fix it. That
has only happened one time. There is already an issue open somewhere about
that bug..

mwethington:
The fix appears to be in NuGet - v2.4 - right? Can we get a blog entry for 2.5? http://reactjs.net/blog/

https://www.nuget.org/packages/React.Web.Mvc4/

mwethington:
OK 2.4 is the latest - 2.5 was for Core stuff changing.

stuartharper01:
Hello. 
Is this fix incorporated into the following NuGet versions please?:
React.Web 2.5.0
React.Web.Mvc4  2.5.0

I ask because I still get the following exception when Load Testing:
JSPool.Exceptions.JsPoolExhaustedException
Could not acquire JavaScript engine within 00:00:05

When I set the React config that stops JS pooling, the Exhausted exception goes away but the performance is adversely affected:
ReactSiteConfiguration.Configuration.SetReuseJavaScriptEngines(false);

Thanks

stuartharper01:
OK thanks. I can see the #270 fix in the source of React.Core\ReactEnvironment.cs.
But unfortunately, I still get the following exception during Load Tests:
JSPool.Exceptions.JsPoolExhaustedException
 Could not acquire JavaScript engine within 00:00:05

huan086:
You'll need to change your configuration to increase your max engines. You're most likely running out of engines in the pool when load gets high and it takes several seconds to execute the script

huan086:
@DaniilSokolyuk could you provide some context to the optimization? What was the bottleneck in @Daniel15 's JSPool implementation?

I'm thinking of implementing my custom high-load JSPool using [AsyncEx AsyncCollection](https://github.com/StephenCleary/AsyncEx/blob/master/doc/AsyncCollection.md). Daniel15's implementation uses BlockingCollection, which blocks threads from the ThreadPool. As load goes up, number of threads that get blocked goes up, available threads to serve non-React views goes down and the whole system slows to a crawl.

DaniilSokolyuk:
@huan086 Hi! I now reimplement JsPool i make it truly async without thread starvation
look
https://github.com/DaniilSokolyuk/ZeroReact.NET/blob/master/src/ZeroReact/JsPool/ZeroJsPool.cs

dustinsoftware:
@DaniilSokolyuk I’d be open to making it so the implementation of engine pooling is swappable, so consumers can choose either jspool or your implementation (or maybe your changes can go upstream into JSPool itself..)

Daniel15:
I'd be happy to accept pull requests to JsPool if you're willing to submit them 😃

DaniilSokolyuk:
ReactJS.Net has sync API,make JsPool async without change  ReactJS.Net not make a sense, and in legacy asp you can’t use await in views, it will work only in asp net core
Ofc we can have async and sync api  with .GetAwaiter().GetResult()

Issue #299, opened 2016-08-08
MSBuild: Only transpile when file has changed


TorbenRahbekKoch:
Since the transpiling does take a considerable amount of time, it would be awesome if the files were only transpiled when changed (or the target file is missing).

It could be as simple as just comparing the timestamps on the original file and the generated file, or something more advanced as storing a hash for the original file. That would probably be overkill, though.

Daniel15:
This is a good idea. The file actually already has a hash (see [Babel.LoadFromFileCache](https://github.com/reactjs/React.NET/blob/master/src/React.Core/Babel.cs#L160)) so the MSBuild task would just need to check the cache before transforming the file.

The best solution is probably to update [Babel.TransformAndSaveFile](https://github.com/reactjs/React.NET/blob/master/src/React.Core/Babel.cs#L348) to check the cache via `LoadFromFileCache` before doing the transformation. If the cached data is valid, we don't need to do the transformation.

Daniel15:
Fixed in #313, thanks @TorbenRahbekKoch!

PR #313, opened 2016-09-03
Use file hash to check for file changes before transpiling (#299)

TorbenRahbekKoch:
In reaction to issue [#299](https://github.com/reactjs/React.NET/issues/299) I have added a hash based check, which then only transpiles the .jsx file if it has actually changed. This shortens build times considerably.

### DIFF
--- a/src/React.Core/Babel.cs
+++ b/src/React.Core/Babel.cs
@@ -351,35 +351,36 @@ namespace React
 		{
 			var outputPath = GetOutputPath(filename);
 			var contents = _fileSystem.ReadAsString(filename);
-            if (MustTranspileFile(contents, outputPath))
-            {
-                var result = TransformWithHeader(filename, contents, null);
+			if (CacheIsValid(contents, outputPath))
+				return outputPath;
+			
+			var result = TransformWithHeader(filename, contents, null);
 
-                var sourceMapPath = GetSourceMapOutputPath(filename);
-                _fileSystem.WriteAsString(outputPath, result.Code);
-                _fileSystem.WriteAsString(sourceMapPath, result.SourceMap == null ? string.Empty : result.SourceMap.ToJson());                
-            }
-		    return outputPath;
+			var sourceMapPath = GetSourceMapOutputPath(filename);
+			_fileSystem.WriteAsString(outputPath, result.Code);
+			_fileSystem.WriteAsString(sourceMapPath, result.SourceMap == null ? string.Empty : result.SourceMap.ToJson());                
+			
+			return outputPath;
 		}
 
-        /// <summary>
-        /// Checks whether an input file (given as inputFileContents) should be transpiled
-        /// by calculating the hash and comparing it to the hash value stored
-        /// in the file given by outputPath. If the outputPath file does not
-        /// exist the input file should always be transpiled.
-        /// </summary>
-        /// <param name="inputFileContents">The contents of the input file.</param>
-        /// <param name="outputPath">The output path of the (possibly previously) generated file.</param>
-        /// <returns>Returns true if the file should be transpiled, false otherwise.</returns>
-	    public virtual bool MustTranspileFile(string inputFileContents, string outputPath)
-	    {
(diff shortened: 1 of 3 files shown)


Daniel15:
Please use tabs rather than spaces. You can use [EditorConfig](https://visualstudiogallery.msdn.microsoft.com/c8bccfe2-650c-4b42-bc5c-845e21f96328) to automatically use the correct indentation configuration in Visual Studio.

Also, please add a unit test testing this functionality :)

src/React.Core/Babel.cs
```diff
@@ -350,12 +350,36 @@ string filename
...
+        /// <summary>
+        /// Checks whether an input file (given as inputFileContents) should be transpiled
+        /// by calculating the hash and comparing it to the hash value stored
+        /// in the file given by outputPath. If the outputPath file does not
+        /// exist the input file should always be transpiled.
+        /// </summary>
+        /// <param name="inputFileContents">The contents of the input file.</param>
+        /// <param name="outputPath">The output path of the (possibly previously) generated file.</param>
+        /// <returns>Returns true if the file should be transpiled, false otherwise.</returns>
+	    public virtual bool MustTranspileFile(string inputFileContents, string outputPath)
```
Daniel15: Would prefer calling this `ValidateCache` and making it return `true` if the cache is valid.

src/React.Core/Babel.cs
```diff
@@ -350,12 +350,36 @@ string filename
 		)
 		{
 			var outputPath = GetOutputPath(filename);
-			var sourceMapPath = GetSourceMapOutputPath(filename);
 			var contents = _fileSystem.ReadAsString(filename);
-			var result = TransformWithHeader(filename, contents, null);
-			_fileSystem.WriteAsString(outputPath, result.Code);
-			_fileSystem.WriteAsString(sourceMapPath, result.SourceMap == null ? string.Empty : result.SourceMap.ToJson());
-			return outputPath;
+            if (MustTranspileFile(contents, outputPath))
```
Daniel15: You could return early here, rather than wrapping the entire thing in the `if`.

src/React.Core/Babel.cs
TorbenRahbekKoch: ValidateCache is rather vague and unspecific - CacheIsValid would be a better name, since it better hints at what state the cache is actually in, when true/false is returned.

And I like flipping the true/false values, since it would enable me to turn around the if in your next comment ;)

Daniel15: Yeah, `CacheIsValid` or `CheckCacheValidity` would be fine.

Also please make it `private`, and it does not need to be `virtual`.

src/React.Core/Babel.cs
TorbenRahbekKoch: The reason I didn't do that in the first place is that NOTs (!) in C# is rather subtle and easy to miss. Renaming the MustTranspileFile to e.g. CacheIsValid would allow me to do that. Good suggestion.

It would give two ```return outputPath;``` statements, though.

TorbenRahbekKoch:
EditorConfig seems like a nice tool. Maybe you should add more .editorconfig files to the relevant parts of the solution ;)

Sorry about the missing unit-test. That was a brain fart on my part.

All in all good suggestions - I will get back to you with some updates ;)

Daniel15:
Ahh, my mistake, I didn't realise this project was missing an editorconfig file! This is the config I normally use: https://github.com/Daniel15/RouteJs/blob/master/.editorconfig. I must have forgotten to copy it across to this repository 😄

Daniel15:
(although I've used tabs for a long time, I'm tempted to switch to spaces since at work we always use two spaces for indentation)

TorbenRahbekKoch:
Thanks for the config ;) I think that, in general, using spaces seems to cause the least hazzle ;)

TorbenRahbekKoch:
I think I'm done now ;)

Daniel15:
Thank you! :D

PR #328, opened 2016-10-05
updated tutorial to remove early reference

mwarger:
Removed a reference to @Html.ReactInitJavaScript() that was added prematurely to the bundling section of the tutorial.  It was not needed until the server-side rendering section.


dustinsoftware:

PR #332, opened 2016-10-10
Update Tutorial to ensure the CommentList and ComentForm declarations…

cheeseytoastie:
… are before CommentBox

As a newbie I just spent half an hour scratching my head why I was getting a null error.


Daniel15:
Thanks!

Issue #353, opened 2016-11-27
ASP.NET Core 1.1 & ReactJS


DaveKP:
Couldn't find a better place to ask - Should ReactJS be working with .net Core 1.1? 

I'm getting :

> ReactNotInitialisedException: ReactJS.NET has not been initialised correctly. Please ensure you have called app.AddReact() and app.UseReact() in your Startup.cs file.

When trying to use server-side rendering.

I can't see anything that I've mis-configured though, based on the tutorial code.

[Startup.cs.txt](https://github.com/reactjs/React.NET/files/614813/Startup.cs.txt)

DaveKP:
Okay, so i managed to get things working. I imported React.Core and React.AspNet from the GitHub repo. In [Initialiser.cs](https://github.com/reactjs/React.NET/blob/master/src/React.Core/Initializer.cs) - the IsReactAssembly() method - for me the assembly names were coming in all lower case, so I changed the string comparisons inside the method. 

I haven't had time to search for why they're coming in lower case, anyone know?

I'm messing around in VS 2017 RC at the moment, so my environment's a mess - but I'll make a pull request when I have time to sort things out. It's bedtime here!

Daniel15:
Hmm strange - I'm running http://dan.cx/socialfeed.htm on Mono 4.x + ASP.NET Core 1.1 + ReactJS.NET 3.0 and it's working fine.

chanan:
Hi,

I think I might have more information on this problem. I have the same exact (I am pretty sure) project (Core 1.1) in VS2015 and in VS2017 RC. I believe the only difference is that in 2017 the project is using the new csproj file format. React.Net works for me in VS2015 but in VS2017 I get the "ReactNotInitialisedException: ReactJS.NET has not been initialised correctly. Please ensure you have called app.AddReact() and app.UseReact() in your Startup.cs file." message.

Daniel15:
@chanan - Hmm, I haven't tested on VS2017 RC yet. Did you add `app.AddReact` and `app.UseReact` to your Startup.cs file like the message says? What's the full error message and stack trace you're seeing?

chanan:
Yes both in VS2015 and VS2017 RC the startup and all the other files are the same. 

[...]

PR #354, opened 2016-11-28
Fixed a problem registering React assemblies with TinyIoC in some instances

DaveKP:
In some instances (unknown cause) RuntimeLibrary names are lower case.
In this case they were failing to register with TinyIoC.

#353


Daniel15:
Thanks!

PR #357, opened 2016-11-30
Demo support for running on .NET Core

chriscamplejohn:
Adds .NET Core as a supported framework
Fixes issue with loading V8 assemblies on .NET Core


dustinsoftware:
Maybe I'm missing something? The tutorial already supports .NET Core, and is working without the changes in this pull request.

Daniel15:
Sorry I completely missed this PR. I don't think these changes are needed any more.

chriscamplejohn:
Yes - correct. This was for V1 I don't believe it is needed in the newer versions of .NET Core.

Issue #338, opened 2016-10-21
VroomJsInitialisationException: Failed to initialise VroomJs


nhannguyen29:
Hi, I'm using macOS Sierra and ASP.NET Core 1.0 for my website. I encountered  this error after adding and configuring React.AspNet 3.0.0 package to my website. Can you help me resolve it? Thank you so much.

Here is the 500 Internal Error I'm getting:

> VroomJsInitialisationException: Failed to initialise VroomJs. This is most likely caused by the native library (libVroomJsNative.so) being out of date or your system lacking a compatible version of V8. Please run Mono with the MONO_LOG_LEVEL=debug environment variable for more debugging information, and refer to the ReactJS.NET Mono documentation at http://reactjs.net/guides/mono.html for further debugging hints.
> 
> More details: Exception has been thrown by the target of an invocation.
> ConstructType
> 
> TinyIoCResolutionException: Unable to resolve type: React.JavaScriptEngineFactory
> ConstructType
> 
> TinyIoCResolutionException: Unable to resolve type: React.ReactEnvironment
> ConstructType

Daniel15:
I haven't tested on Mac OS for a while, and don't have a Mac to test this out on at the moment. You probably need to compile VroomJs. See https://github.com/pauldotknopf/vroomjs-core for instructions.

nhannguyen29:
I tried to follow the instruction on how to set up the VroomJs. However, I encountered this error:

> svn checkout --force http://gyp.googlecode.com/svn/trunk build/gyp \
--revision 1501
>svn: E170013: Unable to connect to a repository at URL 'http://gyp.googlecode.com/svn/trunk'
>svn: E160013: '/svn/trunk' path not found
>Makefile:301: recipe for target 'dependencies' failed
>make: *** [dependencies] Error 1

Seems like the issue #289  is the same as mine. Can you look into this please? Thank you so much.

fujihara:
It will not compile to Mac since the repository http://gyp.googlecode.com/svn/trunk is not available anymore and I could not make VroomJs compile with gyp version greater than 3.17.

Is it possible to run on Mac using javascript engine other than v8?

Taritsyn:
> Is it possible to run on Mac using javascript engine other than v8?

[...]

PR #362, opened 2016-12-15
FIX: VroomJS engine initialize failed on macOS.

holmescn:
#338 mentioned this error, and as the chakra js engine already added, which support `osx-x64` platform, I think the simplest way is to skip the VroomJS engine if it is not available. And the fix is also a defensive programming for this situation.

### DIFF
--- a/src/React.Core/JavaScriptEngineFactory.cs
+++ b/src/React.Core/JavaScriptEngineFactory.cs
@@ -373,7 +373,12 @@ namespace React
 #if NET40
 			jsEngineSwitcher.EngineFactories.AddV8();
 #endif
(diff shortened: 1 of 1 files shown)


Daniel15:
People still using .NET Framework 4.0 or 4.5 on Linux (via Mono) still need to see the VroomJs initialisation error, though. 

Instead of doing it this way, you can modify your app to only add the Chakra engine rather than VroomJs. Just configure the engines before calling `UseReact`:

```
            var engineSwitcher = JsEngineSwitcher.Instance;
            engineSwitcher.EngineFactories
                .AddChakraCore();

            app.UseReact(config =>
```

dustinsoftware:

PR #397, opened 2017-04-16
Fixed to to match current setup instructions

gunnim:
According to
https://reactjs.net/getting-started/tutorial.html
you run services.AddReact() not app.AddReact()

Fixed exception error messages to match


Daniel15:
Thanks!

PR #421, opened 2017-07-02
Upgrade to JSPool 3.0

Daniel15:
Upgrades ReactJS.NET to use JSPool 3.0, and updates all JavaScriptEngineSwitcher dependencies to their latest versions

Also adds an assembly redirect hack for MSBuild so that it can correctly load `JavaScriptEngineSwitcher.Core` when `JavaScriptEngineSwitcher.V8` references an older version. Normally this would be done in `Web.config` or `app.config` as an assembly redirect, but that does not work in MSBuild. The only way to do it in MSBuild is to manually redirect the assembly.

### DIFF
--- a/src/React.Core/JavaScriptEngineFactory.cs
+++ b/src/React.Core/JavaScriptEngineFactory.cs
@@ -223,7 +223,7 @@ namespace React
 		/// Gets a JavaScript engine from the pool.
 		/// </summary>
 		/// <returns>The JavaScript engine</returns>
-		public virtual IJsEngine GetEngine()
+		public virtual PooledJsEngine GetEngine()
 		{
 			EnsureValidState();
 			return _pool.GetEngine();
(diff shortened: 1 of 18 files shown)


src/React.Core/JavaScriptEngineFactory.cs
```diff
@@ -233,13 +233,13 @@ public virtual IJsEngine GetEngine()
...
 		/// <param name="engine">Engine to return</param>
-		public virtual void ReturnEngineToPool(IJsEngine engine)
+		public virtual void ReturnEngineToPool(PooledJsEngine engine)
 		{
 			// This could be called from ReactEnvironment.Dispose if that class is disposed after 
 			// this class. Let's just ignore this if it's disposed.
 			if (!_disposed)
 			{
-				_pool.ReturnEngineToPool(engine);	
+				engine.Dispose();
```
Daniel15: This is probably not needed... The call site could just call `Dispose` on the engine directly.

PR #445, opened 2017-10-16
React 16

dustinsoftware:
The samples still need to be updated. This updates the built-in copy of React. The MVC Core sample was updated to support PropTypes and ES6 classes, both of which are required for React 16.

CommonJS imports are still not supported with built-in babel. I'm not sure how many people still need that functionality since compiling the client + server bundle with webpack ahead of time is a pretty stable alternative. So, it may be worth it to just deprecate on-the-fly JS compiler and move ahead with webpack-only builds.


damanptyltd:
This is great Dustin, would love to see this merged into a release.

dustinsoftware:
@Daniel15 this is almost ready to go. I just have to finish getting the webpack sample working before this should be merged, everything else is ready for review

dustinsoftware:
@Daniel15 this should be good to go now.

Daniel15:
Thank you @dustinsoftware! I'll try to push out a release this weekend.

Daniel15:
Released this today as version 3.2. Thanks!

Issue #341, opened 2016-10-26
"BYO React" page no longer exists


wub:
When I tried to load my own React, I got this error:

```
React has not been loaded correctly.
Please expose your version of React as global variables named 'React', 'ReactDOM' and 'ReactDOMServer', or enable the 'LoadReact' configuration option to use the built-in version of React.
See http://reactjs.net/guides/byo-react.html for more information.
```

I ended up loading the built-in version, but I'm more concerned that the documentation no longer exists to fix it (at this URL at least!)

https://github.com/reactjs/React.NET/blob/e12020b1f99f99d71f22a92401a2e23913e891bc/src/React.Core/JavaScriptEngineFactory.cs#L185

Daniel15:
Oops, I don't think I ever actually wrote the documentation! I think that was just a placeholder URL 😛

szarouski:
@Daniel15 I can customize react version by applying changes like shown here (webpack file):
https://github.com/reactjs/React.NET/commit/27df77f84f2ea95994452b82f0d55c48493ba2df

My only problem is that things don't work when I set .SetLoadReact(false), but without it I can see custom version of react. Do you think there might be any concerns regarding not using `.SetLoadReact(false)`?

Thank you.

wub:
https://github.com/reactjs/React.NET/pull/454 🤙

PR #454, opened 2017-11-03
Removed dead (not-yet-alive?) byo-react link

JoshuaKGoldberg:
See https://github.com/reactjs/React.NET/issues/341.

If the page gets created, it could be added in.


Daniel15:
Thanks... I guess I never created the page.

PR #459, opened 2017-11-04
Removed engine-specific logic from configuration and the engine factory

### LINKED ISSUE #456: Does JavaScriptEngineFactory.cs really need a dependency on V8 or Chakra?
Deploying multiple JS engine DLLs is a pain and extra time taken during server deployments. V8 in particular has a few DLL dependencies and can become DLL hell.

In setting up ReactJS.NET with a server, I ended up wanting to both set up my own container registrations and only ever use the Chakra engine. Here's the code it ended up with:

```csharp
var container = AssemblyRegistration.Container;
var fileSystem = new SimpleFileSystem();

var msIeEngineFactory = new MsieJsEngineFactory();
JsEngineSwitcher.Instance.EngineFactories.Add(msIeEngineFactory);
JsEngineSwitcher.Instance.DefaultEngineName = "MsieJsEngine";

var engineFactory = new JavaScriptEngineFactory(
    JsEngineSwitcher.Instance,
    ReactSiteConfiguration.Configuration,
    fileSystem);

container.Register<IReactSiteConfiguration>((c, o) => ReactSiteConfiguration.Configuration);
container.Register<ICache, NullCache>();
container.Register<IFileCacheHash, FileCacheHash>().AsPerRequestSingleton();
container.Register<FileCacheHash, FileCacheHash>().AsPerRequestSingleton();
container.Register<IFileSystem, SimpleFileSystem>(fileSystem);
container.Register<JsEngineSwitcher>((c, o) => JsEngineSwitcher.Instance);
container.Register<IJavaScriptEngineFactory>(engineFactory);
container.Register<IReactEnvironment, ReactEnvironment>().AsPerRequestSingleton();
container.Register<ReactEnvironment, ReactEnvironment>().AsPerRequestSingleton();

    .SetLoadBabel(false)
    .AddScriptWithoutTransform("local/Wat.js");
```

The `JavaScriptEngineFactory` call loads [JavaScriptEngineFactory.cs](https://github.com/reactjs/React.NET/blob/master/src/React.Core/JavaScriptEngineFactory.cs), which has a dependency on `JavaScriptEngineSwitcher.V8` or `JavaScriptEngineSwitcher.ChakraCore`:

> ```csharp
> #if NET40
> using JavaScriptEngineSwitcher.V8;
> #else
> using JavaScriptEngineSwitcher.ChakraCore;
> #endif
> ```

That means if you want to load that file, you have to load whichever engine switcher DLL is tied to your .NET version. Eww! In order to avoid having to figure out V8 deployments, I ended up copy & pasting `JavaScriptEngineFactory` into a new file and removing the references to `V8`. Forking is bad and I feel bad.

A better solution IMO would be to only load the DLLs deemed necessary. How about `JavaScriptEngineFactory` take in some sort of new `IJavaScriptEngineProvider`(s) via dependency injection? That way we can skip loading the V8 or Chakra DLLs unless necessary. It would also help remove some of the logic for switching between the two in the class itself, which as it stands now seems like a weird responsibility for the class to have.

JoshuaKGoldberg:
The core DLL shouldn't have dependencies on specific engine DLLs such as ChakraCore or V8, regardless of .NET version. This change removes `AllowMsieEngine`/`allowMsie` from `ReactSiteConfiguration.cs` and any Chakra, V8, and Vroom-specific code from `JavaScriptEngineFactory`.

Fixes #456.


JoshuaKGoldberg:
Users would now be required to set the default factory on their own. In a subsequent change, I'd like to add that logic back in with just the JavaScriptEngineSwitcher APIs. @Daniel15 what are your thoughts on where that should go, if anywhere? Collecting the comments from #456:
* A new `IJavaScriptEngineChooser` Func that returns the name of the preferred engine
* Conditionally set it based on .NET version _(or this as the default `IJavaScriptEngineChooser`?)_
* Nowhere - make them do it on their own! (breaking change?)

Taritsyn:
> * A new `IJavaScriptEngineChooser` Func that returns the name of the preferred engine

I vote for this variant. But I'm not sure about the advisability of adding a new interface - `IJavaScriptEngineChooser`. There would be more appropriate the configuration property is responsible for setting engine name (similar approach I use in the [Bundle Transformer](https://github.com/Taritsyn/BundleTransformer)). JavaScript Engine Switcher initially has an infrastructure for such things (see the [“Registration of JS engines”](https://github.com/Taritsyn/JavaScriptEngineSwitcher/wiki/Registration-of-JS-engines) and [“Creating instances of JS engines”](https://github.com/Taritsyn/JavaScriptEngineSwitcher/wiki/Creating-instances-of-JS-engines) sections of documentation).

src/React.Core/IReactSiteConfiguration.cs
```diff
@@ -114,16 +114,14 @@ public interface IReactSiteConfiguration
 		IReactSiteConfiguration SetMaxUsagesPerEngine(int? maxUsagesPerEngine);
 
 		/// <summary>
-		/// Gets or sets whether the MSIE engine should be used if V8 is unavailable.
+		/// Gets or sets the name of the preferred JavaScript engine.
 		/// </summary>
-		[Obsolete("This should be managed in the JavaScriptEngineSwitcher configuration instead")]
-		bool AllowMsieEngine { get; set; }
+		string PreferredEngineName { get; set; }
```
Daniel15: Can we just use the JavaScriptEnginerSwitcher property for this? I think it has a preferred/default engine setting too.

Daniel15:
Thanks, this looks pretty good to me.

The build failure is a tricky one. It comes from the MSBuild task. The issue is that the MSBuild task doesn't run any of your app code, so you have no chance to configure a JS engine. The solution there might be to enforce a particular JS engine config for the MSBuild task, which you can do in [`MSBuildHost.Initialize`](https://github.com/reactjs/React.NET/blob/master/src/React.MSBuild/MSBuildHost.cs#L38:23). Maybe just add ClearScriptV8 + ChakraCore there?

dustinsoftware:
@JoshuaKGoldberg it looks like this PR is getting stale, so I opened a new one that addresses these issues. #501

JoshuaKGoldberg:
Thanks! I'll close this one :) ran out of time to investigate..

PR #461, opened 2017-11-04
Upgrade React.Router to ASP.NET Core 2.0

Daniel15:
Upgrades `React.Router` to ASP.NET 2.0.

Also splits ASP.NET MVC 4 support out into a separate `React.Router.Mvc4` project. This is required as you can use both ASP.NET Core and "classic" ASP.NET from .NET Framework, so simply discriminating based on framework version is not sufficient, and it actually needs to be two separate packages.

cc @dustinsoftware

### DIFF
--- a/src/React.Router.Mvc4/React.Router.Mvc4.csproj
+++ b/src/React.Router.Mvc4/React.Router.Mvc4.csproj
@@ -8,7 +8,7 @@
 		<TargetFrameworks>net451</TargetFrameworks>
 		<GenerateDocumentationFile>true</GenerateDocumentationFile>
 		<AssemblyName>React.Router.Mvc4</AssemblyName>
-	  <DefineConstants>$(DefineConstants);LEGACYASPNET</DefineConstants>
+		<DefineConstants>$(DefineConstants);LEGACYASPNET</DefineConstants>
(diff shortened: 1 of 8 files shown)


src/React.Router.Mvc4/React.Router.Mvc4.csproj
```diff
@@ -0,0 +1,47 @@
...
+		<PackageId>React.Router.Mvc4</PackageId>
+		<PackageTags>asp.net;mvc;asp;javascript;js;react;facebook;reactjs;babel;router;react router</PackageTags>
+		<PackageIconUrl>http://reactjs.net/img/logo_64.png</PackageIconUrl>
+		<PackageProjectUrl>http://reactjs.net/</PackageProjectUrl>
+		<PackageLicenseUrl>https://github.com/reactjs/React.NET#licence</PackageLicenseUrl>
+		<GenerateAssemblyInfo>false</GenerateAssemblyInfo>
+	</PropertyGroup>
+
+	<ItemGroup>
+    <Compile Include="..\React.Router\*.cs" />
```
Daniel15: I need to fix the indentation in this file.

dustinsoftware:
LGTM. Would be cool to have a sample as well to show off how it is used...
I can put one together if you don't have time

Daniel15:
Unfortunately I don't have time to put a sample together :(

PR #473, opened 2017-11-12
Add optional error boundary support

dustinsoftware:
I updated a sample so you can see how this works. When specifying an error handler, then errors will be caught during server-render (including those not handled by error boundaries).

Some tests still need to be added, but I wanted to get the discussion started about this approach.


dustinsoftware:
Ok @Daniel15 this is ready for review

dustinsoftware:
@Daniel15 would you have time to take a look at this PR soon? Are you OK with the overall direction?

src/React.Core/ReactComponent.cs
```diff
@@ -120,39 +120,42 @@ public virtual string RenderHtml(bool renderContainerOnly = false, bool renderSe
...
 						: string.Format("ReactDOMServer.renderToString({0})", GetComponentInitialiser());
 					html = _environment.Execute<string>(reactRenderCommand);
 				}
-
-				string attributes = string.Format("id=\"{0}\"", ContainerId);
-				if (!string.IsNullOrEmpty(ContainerClass))
+				catch (JsRuntimeException ex)
 				{
-					attributes += string.Format(" class=\"{0}\"", ContainerClass);
+					if (_configuration.ExceptionHandler == null) {
```
Daniel15: I wonder if we should also allow setting an exception handler per component (eg. as an argument to this method). Hmm.

Daniel15: Instead of checking if this is `null`, what about making it the default handler? That way, `_configuration.ExceptionHandler` would never be null.

src/React.Core/ReactComponent.cs
```diff
@@ -120,39 +120,42 @@ public virtual string RenderHtml(bool renderContainerOnly = false, bool renderSe
...
+						throw new ReactServerRenderingException(string.Format(
+							"Error while rendering \"{0}\" to \"{2}\": {1}",
+							ComponentName,
+							ex.Message,
+							ContainerId
+						));
+					}
+					_configuration.ExceptionHandler(ex);
```
Daniel15: Might be worth passing `ComponentName` and `ContainerId` too.

dustinsoftware:
Thanks for the review, I'll push an update soon :)

dustinsoftware:
@Daniel15 this should be good to go, want to give it another look?

Issue #472, opened 2017-11-11
React.Router has a flaky test


Daniel15:
The build seems to randomly fail on `React.Tests.Router.HtmlHelperExtensionsTest.ShouldRunCustomContextHandler`:
```
[xUnit.net 00:00:00.2936128]   Discovering: React.Tests
  [xUnit.net 00:00:00.7217786]   Discovered:  React.Tests
  [xUnit.net 00:00:00.9747520]   Starting:    React.Tests
  [xUnit.net 00:00:02.4140782]     React.Tests.Router.HtmlHelperExtensionsTest.ShouldRunCustomContextHandler [FAIL]
  [xUnit.net 00:00:02.4150445]       System.NullReferenceException : Object reference not set to an instance of an object.
  [xUnit.net 00:00:02.4157439]       Stack Trace:
  [xUnit.net 00:00:02.4172806]            at React.Router.HtmlHelperExtensions.ReactRouterWithContext[T](HtmlHelper htmlHelper, String componentName, T props, String path, String htmlTag, String containerId, Boolean clientOnly, Boolean serverOnly, String containerClass, Action`2 contextHandler)
  [xUnit.net 00:00:02.4174844]         Router\HtmlHelperExtensionsTest.cs(204,0): at React.Tests.Router.HtmlHelperExtensionsTest.ShouldRunCustomContextHandler()
  [xUnit.net 00:00:02.4358859]   Finished:    React.Tests
  Failed   React.Tests.Router.HtmlHelperExtensionsTest.ShouldRunCustomContextHandler
EXEC : error Message:  [C:\projects\react-net\build.proj]
   System.NullReferenceException : Object reference not set to an instance of an object.
  Test Run Failed.
  Stack Trace:
     at React.Router.HtmlHelperExtensions.ReactRouterWithContext[T](HtmlHelper htmlHelper, String componentName, T props, String path, String htmlTag, String containerId, Boolean clientOnly, Boolean serverOnly, String containerClass, Action`2 contextHandler)
     at React.Tests.Router.HtmlHelperExtensionsTest.ShouldRunCustomContextHandler()
```

cc @gunnim

gunnim:
https://github.com/reactjs/React.NET/pull/475 seems to take care of the issue, i'm no longer able to reproduce the bug 👍

Daniel15:
Thank you for fixing it, @dustinsoftware :D

PR #474, opened 2017-11-13
Fix xUnit tests in VS2017

dustinsoftware:
Due to an xUnit bug, tests would attempt to run, but then fail
immediately. Updating the packages fixes the issue.

https://github.com/Microsoft/vstest/issues/811


Daniel15:
I was hoping this would fix #472, but it still seems flaky (based on the AppVeyor failure). `React.Tests.Router.HtmlHelperExtensionsTest` fails sporadically. :(

dustinsoftware:
I found the fix and will open another PR in just a minute..... :)

PR #475, opened 2017-11-13
Fix flakey tests

dustinsoftware:
This is caused by ConfigureMockEnvironment setting up the static
property Environment. Since tests run in parallel, there is a race
between which tests that set up different fake environments, causing
occasional failures. Disabling parallelization fixes the issue, and
tests run so fast that this should be ok.


Daniel15:
Thanks! I think `HtmlHelperExtensions` is the only part of ReactJS.NET that doesn't inject dependencies into its constructor (since it consists entirely of static extension methods) and thus relies on some global environment. I wonder if that could be changed/fixed.

Issue #476, opened 2017-11-17
Add react router sample


dustinsoftware:
I'm planning on a new sample for react router, based on the CoreMvc + webpack samples that exist already.

Daniel15:
Thanks! 😃

PR #479, opened 2017-11-23
Add react router sample

dustinsoftware:


dustinsoftware:
Fixes #476

src/React.Router/HtmlHelperExtensions.cs
```diff
@@ -123,7 +123,7 @@ var reactComponent
 
 				var executionResult = reactComponent.RenderRouterWithContext(clientOnly, serverOnly);
 
-				if (executionResult.Context?.status != null)
+				if (executionResult.Context?.status != null || executionResult.Context?.url != null)
```
Daniel15: Why is this change needed?

src/React.Router/SetServerResponse.cs
```diff
@@ -36,7 +36,7 @@ public static class SetServerResponse
 		/// <param name="Response">The response object to use.</param>
 		public static void ModifyResponse(RoutingContext context, HttpResponse Response)
 		{
-			var statusCode = context.status.Value;
+			var statusCode = context.status ?? 302;
```
Daniel15: Why is this change needed?

src/React.Router/HtmlHelperExtensions.cs
dustinsoftware: It is possible for a redirect to be returned [without a status code defined](https://github.com/ReactTraining/react-router/blob/master/packages/react-router/docs/api/Redirect.md)

src/React.Router/SetServerResponse.cs
dustinsoftware: If a status code is omitted we can't know if the author intended a permanent or temporary redirect, therefore it's safest to assume temporary

Daniel15:
Thanks!

PR #492, opened 2018-01-03
Propose using weback export settings over expose-loader

starwed:
I found it was a bit less boilerplate to expose components this way, rather than by using expose-loader.

Not 100% sure this is the best way to integrate into existing docs, but figured I'd offer it as a quick suggestion.  (Especially, might need some reworking to use with separate client/server bundles.)

### DIFF
--- a/site/jekyll/guides/webpack.md
+++ b/site/jekyll/guides/webpack.md
@@ -25,6 +25,16 @@ module.exports = {
 var Components = require('expose?Components!./components');
 ```
 
+An alternative to using expose-loader is to use the `library` property of the output configuration.
+``` javascript
+ output: {
+    path: path.join(__dirname, 'build'),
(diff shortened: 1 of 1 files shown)


dustinsoftware:
Thanks for the suggestion. I'm not quite following how this is more simple than just one call to `require('expose-loader?Components!./components)'. As long as all of your components are available in that object, you should be good to go. Per the [docs] it looks like this webpack feature is intended for shipping libraries that need to run in multiple environments and isn't the best fit here.

Daniel15:
Won't you get lots of code duplication if you export each component as a separate library?

dustinsoftware:
Hi there, it looks like this PR is getting stale and is unfortunately not mergeable as is. I'm going to close it, but feel free to revisit it in the future if you'd like!

PR #493, opened 2018-01-04
Docs: Update expose-loader syntax for latest Webpack

gazab:
When using expose-loader with later versions of Webpack you need to keep -loader in the name like this:
`var Components = require('expose-loader?Components!./components'); `
instead of
`var Components = require('expose?Components!./components');`

This PR updates the docs with this change since most people needing this won't use Webpack 1.


dustinsoftware:
lgtm

Issue #334, opened 2016-10-11
Can't add attributes to `ReactInitJavaScript` script tag


wub:
We'd like to be able to add extra attributes to the `ReactInitJavaScript` tag:

*Current output:*
`<script>ReactStuff()</script>`

*Possible desired output:*
`<script async foo="bar" nonce="Fo0/B4r">ReactStuff()</script>`

If it's a simple case of adding another parameter to the helper, and adding more to the TagBuilder, I'm happy to take a stab at this.

Daniel15:
What's an example of attributes you'd like to add, and for what purpose?

wub:
Hey @Daniel15, I managed to just reuse some code from inside ReactInitJavaScript and made my own helper. I'll  post it here tomorrow when I'm back at work.

I needed to add both "async" and "nonce" (we're slowly moving to CSP).

Cheers!

Daniel15:
I could be wrong, but I don't think "async" would make a difference since it's an inline script.

wub:
You are not wrong :0)

Issue #482, opened 2017-11-27
Add better content security policy support


dustinsoftware:
Instead of needing to specify `unsafe-inline`, allow specifying a "nonce" to the rendered script tags. If this nonce matches the value in the CSP header, the script will be allowed to run.

[MDN docs](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/script-src)

dustinsoftware:
I'm planning on opening a PR for this one without objection

PR #496, opened 2018-01-13
Add support for script nonce attributes

### LINKED ISSUE #430: Custom attributes in `ReactInitJavaScript` script tag
Same as the now closed issue #334 

Like the aforementioned post, I'm about to implement CSP on a site using ReactJS.NET.
For the inline script created by ReactInitJavaScript to work i need to add a nonce attribute to the script tag with a value that matches one in my CSP headers.

Currently I have created my own helper, similar in functionality to ReactInitJavaScript. 
However, seeing as this configuration might become more and more common as time goes on, I think it might make sense to integrate into the project. 

Do you agree? 
And yes I'd be willing to submit a pull request :)

dustinsoftware:
Fixes #430, #482, #334

### DIFF
--- a/src/React.Core/IReactEnvironment.cs
+++ b/src/React.Core/IReactEnvironment.cs
@@ -1,4 +1,4 @@
-﻿/*
+/*
  *  Copyright (c) 2014-Present, Facebook, Inc.
  *  All rights reserved.
  *
@@ -7,7 +7,6 @@
  *  of patent rights can be found in the PATENTS file in the same directory.
  */
 
-using System;
 
 namespace React
 {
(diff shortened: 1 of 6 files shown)


dustinsoftware:
@Daniel15 this is good to go, but I'd like another set of eyes to give it a quick look before it gets merged 😄

dustinsoftware:
Merging this so it doesn't get stale, let me know if needs modification before shipping. Tested on Mvc4 and CoreMvc.

src/React.Core/IReactEnvironment.cs
```diff
@@ -110,5 +109,10 @@ public interface IReactEnvironment
 		/// Returns the currently held JS engine to the pool. (no-op if engine pooling is disabled)
 		/// </summary>
 		void ReturnEngineToPool();
+
+		/// <summary>
+		/// Gets the site-wide configuration.
+		/// </summary>
+		IReactSiteConfiguration Configuration { get; }
```
Daniel15: Isn't there already a getter for this? `ReactSiteConfiguration.Configuration` or `.Current` or something like that?

dustinsoftware:
I can check later, the prop already existed on the implementation but not
the interface.. don’t remember seeing another one I could use

PR #497, opened 2018-01-15
Improve server side only rendering

### LINKED ISSUE #340: Html.ReactInitJavaScript() outputs server only components in initialization code
Using `@Html.React("Component", ..., serverOnly: true)` still adds the `ReactDOM.render()` calls to the `@Html.ReactInitJavaScript()` output which breaks pages using server only components. (It also wraps the output in an unnecessary container element)

Looking through the code there seems to be missing a bool in the ReactEnvironment `CreateComponent` method that sets the component to be server only and thus skipped when outputting the init code for all components

gazab:
This PR fixes two issues with rendering server side ONLY components:
- Component containers are no longer rendered
- @Html.ReactInitJavaScript() no longer outputs JavaScript hydrating the component

For example, consider the following code:
`@Html.React("HelloWorld", Model, serverOnly: true)`

Before the fix this resulted in the following HTML:
`<div id="react-container123"><h1>Hello world!</h1></div>`

And Html.ReactInitJavaScript() rendered:
`<script>ReactDOM.hydrate(React.createElement(HelloWorld, {"name":"example"}), document.getElementById("react-container123"));</script>`

despite the serverOnly = true argument

After this fix the component will be rendered like this:
`<h1>Hello world!</h1>`
and ReactInitJavaScript() won't render anything at all for this component.

This should fix #340


gazab:
I'll rebase tomorrow

gazab:
@dustinsoftware fixed! If you wonder why I changed some of the test assertions it's because they made more sense that way. Passing true to either clientOnly or serverOnly, but not both at the same time. 

Ideally, maybe those parameters should be joined to an enum or something instead. I.e enum RenderMode: ClientAndServer, ClientOnly, ServerOnly.

I didn't want to introduce too big changes though so I left it like it was.

dustinsoftware:
Merged

PR #498, opened 2018-01-19
Include Views web.config transform in mvc4 package

gunnim:
When you install the mvc4 package currently you are missing the react.router namespace.
This means you cannot easily access the HtmlHelper react.router extension method.

On a side note, what was the reasoning behind the mvc4 package again? With this change it seems the react.router package and mvc4 are identical, should we remove the content folder from react.router ? and should the mvc4 package not include react.router as a dependency and lose it's duplicate code files?


gunnim:
Duplicate

PR #499, opened 2018-01-19
React.router.mvc4 package fixes

gunnim:
When you install the mvc4 package currently you are missing the react.router namespace.
This means you cannot easily access the HtmlHelper react.router extension method.
You also need React.Web.Mvc4 to use Html.ReactInitJavaScript so a dependency should be included as well right?

On a side note, what was the reasoning behind the mvc4 package again? With this change it seems the react.router package and mvc4 are identical, should we remove the content folder from react.router ? and should the mvc4 package not include react.router as a dependency and lose it's duplicate code files?


dustinsoftware:
The MVC4 project targets the full ASP.NET MVC framework, as far as I can tell there is no good way to combine the NET Core version and NET Standard other than the way it is currently set up.

gunnim:
reference `React.Web.Mvc4` from `React.Router.csproj` ?
React.Router w/o React.Router.Mvc4 might be used by .NET Core.
F.x. in the sample react.router .net core project we would end up including asp.net v5 code.
Instead I propose we add React.AspNet as a dependency for React.Router, that would include the correct IHtmlHelper extension methods for .net core.

What we also need is a transform for the .net core _ViewImports file both for React.AspNet and React.Router, but i can't find a simple way to do it.

Daniel15:
> as far as I can tell there is no good way to combine the NET Core version and NET Standard other than the way it is currently set up.

I think that's right. The full .NET Framework supports both the full ASP.NET as well as .NET Core, and you can't have one single project or assembly that targets both, so they need to be separate.

dustinsoftware:
Merged

Issue #463, opened 2017-11-06
Standardize source linting: spaces vs tabs; common copyright header


JoshuaKGoldberg:
_(this issue is motivated 50% by code quality, 50% by me familiarizing myself with the code base)_

I couldn't help but notice a few files with spaces intermixed with the tabs :smile:. There are also...

* A few different copyright header variations
* At least one file (`Babel.cs`) with nonstandard line endings.
* Trailing whitespace occasionally

Daniel15:
It's supposed to use tabs in most places, and spaces in .json files. See the editorconfig: https://github.com/reactjs/React.NET/blob/master/.editorconfig

The copyright header is supposed to be 2014-Present, someone updated it via a pull request a while back. Maybe they missed some of the files

dustinsoftware:
Line endings should be fixed now

PR #500, opened 2018-01-20
Normalize newlines

dustinsoftware:
Fortunately only a small set of files were affected - this adds a `gitattributes` file so the line endings stay consistent from now on.

Fixes the major issue identified in #463

To fix any merge conflicts introduced by this PR, a manual merge is needed.

1. Cherry pick commit `0bfbb0e ` to the affected branch
1. Delete `.git\index`
1. Run `git reset`
1. Commit the files

The merge conflicts should be resolved now 😄 ✅


Daniel15:
Thanks! I guess much of the mismatching in my code was from when I was using a MacBook Pro as my work laptop, and switching between MacOS and Windows.

Why normalize with LF though, rather than CR+LF?

dustinsoftware:
Fewer files changed.. plus LF is the default for * text=auto.

Mostly I just wanted the change to be as non-disruptive as possible :)

PR #501, opened 2018-01-20
Removed engine-specific logic from configuration and the engine factory

dustinsoftware:
Fixes merge conflicts in #459 and the error with the msbuild compilation step. Still need to re-test some samples, as they have changed since the PR was initially opened (the build is failing because of broken samples)


Daniel15:
I'll try to take a look at this soon. Does it break the out-of-the-box experience, requiring more manual configuration to get started?

dustinsoftware:
It looks like that wasn't considered in the original pull request.. I'll see if there's a way to preserve the original behavior with this work.

dustinsoftware:
I got close but I don't have time to finish this one. Going to focus on making samples and documentation more accessible for now.

PR #509, opened 2018-02-19
Use NuGet reference for tutorial code

dustinsoftware:
This was an oversight during the original .NET Core 2.0 port


Daniel15:
Thanks! :)

PR #514, opened 2018-03-01
Replace "path" with "location" for SSR + React Router.

ChadBurggraf:
According to the documentation, as well as my testing,
React Router expects the URL supplied to `StaticRouter`
to be provided via the `location` prop name.

React.NET currently supplies this value via the
`path` prop name.

### DIFF
--- a/src/React.Router/ReactRouterComponent.cs
+++ b/src/React.Router/ReactRouterComponent.cs
@@ -69,7 +69,7 @@ namespace React.Router
 		{
 			return string.Format(
 				@"React.createElement
-					({0}, Object.assign({1}, {{ path: '{2}', context: context }}))",
+					({0}, Object.assign({1}, {{ location: '{2}', context: context }}))",
 				ComponentName,
 				_serializedProps,
 				_path


Daniel15:
@dunnim or @dustinsoftware could you please take a look at this? I don't have much experience with ReactRouter. How was this working before? Did it change in a newer Router version?

Daniel15:
@gunnim or @dustinsoftware could you please take a look at this? I don't have much experience with ReactRouter. How was this working before? Did it change in a newer Router version?

dustinsoftware:
This change makes the React.NET props more consistent with React Router. Previous use in our sample:

`<StaticRouter context={this.props.context} location={this.props.path}>`

It is a trivial change for any current consumers of React.Router, and it's worth merging for better consistency.

gunnim:
The idea is that you always wrap the StaticRouter, allowing you to render children and target the props shown in the example by dustinsoftware 
My original choice probably had to do with the fact that the location prop for the router also supports location objects which we are never passing down on the server side.

I think I agree this change would be more consistent but I think it's also trivial to continue to support those using the original implementation by simply passing down a location prop as well as the "deprecated" path prop.

dustinsoftware:
Thanks gunnim. I think we should still stick with this change, passing down both may trigger prop-type warnings (such as [forbidExtraProps](https://github.com/airbnb/prop-types/blob/85e584ecb36583788ab9295faa51e85a4bdbd0ff/test/forbidExtraProps.jsx)), and migrating should be pretty painless. Since React Router support is pretty new the risk should be pretty low that someone will get blindsided by this change (it gets called out in the release notes)

gunnim:
True, good points :)

PR #522, opened 2018-03-24
Avoid memory allocation and improve performance, part 1

DaniilSokolyuk:
Hi @Daniel15, for more than a year we have been using this library, and recently noticed problems with allocations and object hits in Large Object Heap if the component is large or a large props is serialized

Goals
1. Native implementation for IHtmlString without recopy and allocating new strings
2. Reuse everything
3. ComponentExistsChecks configuration property (default true), for disable "EnsureComponentExists" (avoid extra engine call, in production for example)

Branch with banchmarks here https://github.com/DaniilSokolyuk/React.NET/tree/bench
Note the difference between the nano seconds and microseconds

**Before, Small objects**

|              Method |     Mean |     Error |    StdDev |    Gen 0 |  Gen 1 | Allocated |
|-------------------- |---------:|----------:|----------:|---------:|-------:|----------:|
|     CreateComponent | 12.97 us | 0.0954 us | 0.0845 us |   1.8463 | 0.5035 |   10.9 KB |
|               React | 17.19 us | 0.2015 us | 0.1787 us |   7.8430 | 1.1597 |  43.81 KB |
| ReactInitJavaScript | 75.81 us | 0.5654 us | 0.5012 us | 146.9727 |      - | 602.56 KB |

**After, Small objects**

|              Method |        Mean |     Error |    StdDev |  Gen 0 |  Gen 1 | Allocated |
|-------------------- |------------:|----------:|----------:|-------:|-------:|----------:|
|     CreateComponent |    541.4 ns |  10.55 ns |  10.83 ns | 0.0496 | 0.0143 |     312 B |
|               React | 12,973.0 ns | 119.90 ns | 100.12 ns | 1.8005 | 0.4883 |   11320 B |
| ReactInitJavaScript | 11,401.2 ns |  72.78 ns |  64.52 ns | 0.1831 |      - |     792 B |

**Before, Large objects**

|              Method |      Mean |     Error |    StdDev |    Gen 0 |    Gen 1 |   Gen 2 |  Allocated |
|-------------------- |----------:|----------:|----------:|---------:|---------:|--------:|-----------:|
|     CreateComponent |  35.76 us | 0.2885 us | 0.2558 us |   5.6763 |   1.5259 |       - |    34.8 KB |
|               React |  48.16 us | 0.4108 us | 0.3843 us |  24.7803 |   3.5400 |       - |  139.08 KB |
| ReactInitJavaScript | 417.16 us | 4.9953 us | 4.6726 us | 356.4453 | 142.5781 | 71.2891 | 1980.55 KB |

**After, Large objects**

|              Method |        Mean |     Error |     StdDev |   Gen 0 |  Gen 1 | Allocated |
|-------------------- |------------:|----------:|-----------:|--------:|-------:|----------:|
|     CreateComponent |    551.0 ns |  10.95 ns |   9.710 ns |  0.0496 | 0.0143 |     312 B |
|               React | 36,337.2 ns | 130.36 ns | 121.939 ns |  5.0049 | 1.3428 |   31585 B |
| ReactInitJavaScript | 51,691.7 ns | 766.71 ns | 717.185 ns | 24.8413 | 0.5493 |  104398 B |


dustinsoftware:
Thanks for starting this conversation. It looks like there are a few breaking changes (like making component disposable), which could be a problem. This PR is also quite large, it would be better to split up optimization’s into smaller PRs so we can discuss each change :) 

Otherwise as is it’s going to be a bit before we can work through this one.

DaniilSokolyuk:
@dustinsoftware i think disposable component is the smallest problem :)
I do not think that there are so many changes here, backward compatibility is also completely left, I do not consider it necessary to split pr

Daniel15:
This is probably the largest pull request we've ever received :) Thanks for workijng on it.

There's a lot of changes that change `#if NET40` to `#if NET40 || NET451`, is it possible to split those out into a separate pull request? As @dustinsoftware said, it's much easier to review multiple small PRs instead of one giant one.

DaniilSokolyuk:
@Daniel15 Thanks you, splitted, after this PR I will create another requester

DaniilSokolyuk:
@Daniel15, @dustinsoftware what about this PR? I have already split this request and prepared the
[second part](https://github.com/DaniilSokolyuk/React.NET/tree/avoid-memory-allocations-part2)
We are already using these changes and significantly reduced the hits in LOH and FullGC

dustinsoftware:
When you’re ready, please open some smaller pull requests so that we can
review and merge those changes individually :)

DaniilSokolyuk:
@dustinsoftware i am already split this PR, there are only 227 new lines here, and about 60% this lines is replace string.format to TextWriter.Write

src/React.AspNet/HtmlHelperExtensions.cs
```diff
@@ -144,55 +149,34 @@ private static IReactEnvironment Environment
...
-			{
-				InnerHtml = script,
-			};
-
+			writer.Write("<script");
 			if (Environment.Configuration.ScriptNonceProvider != null)
 			{
-				tag.Attributes.Add("nonce", Environment.Configuration.ScriptNonceProvider());
+				writer.Write(" nonce=\"");
+				writer.Write(Environment.Configuration.ScriptNonceProvider());
```
dustinsoftware: I see that we're manually constructing HTML tags in existing code.. as long as we're not passing user-generated data to these methods, this should be OK (although it would be nice to sanitize the data going into these attributes just in case)

src/React.Core/React.Core.csproj
```diff
@@ -37,7 +37,7 @@
     <PackageReference Include="JavaScriptEngineSwitcher.Msie" Version="2.4.9" />
     <PackageReference Include="JSPool" Version="3.0.1" />
     <PackageReference Include="MsieJavaScriptEngine" Version="2.2.2" />
-    <PackageReference Include="Newtonsoft.Json" Version="9.0.1" />
+    <PackageReference Include="Newtonsoft.Json" Version="10.0.3" />
```
dustinsoftware: Is bumping `Json.NET` required for this PR?

src/React.Core/ReactComponent.cs
```diff
@@ -114,31 +101,61 @@ public ReactComponent(IReactEnvironment environment, IReactSiteConfiguration con
 		/// <param name="renderServerOnly">Only renders the common HTML mark up and not any React specific data attributes. Used for server-side only rendering.</param>
 		/// <param name="exceptionHandler">A custom exception handler that will be called if a component throws during a render. Args: (Exception ex, string componentName, string containerId)</param>
 		/// <returns>HTML</returns>
-		public virtual string RenderHtml(bool renderContainerOnly = false, bool renderServerOnly = false, Action<Exception, string, string> exceptionHandler = null)
+		public string RenderHtml(bool renderContainerOnly = false, bool renderServerOnly = false, Action<Exception, string, string> exceptionHandler = null)
```
dustinsoftware: Removing `virtual` is a breaking change, can this be avoided?

src/React.Core/ReactComponent.cs
```diff
@@ -204,39 +236,44 @@ protected virtual void EnsureComponentExists()
...
+		/// Write serialized Props to writer
 		/// </summary>
-		/// <param name="componentName"></param>
-		internal static void EnsureComponentNameValid(string componentName)
+		/// <param name="writer">The <see cref="T:System.IO.TextWriter" /> to which the content is written</param>
+		protected void WriteSerializedProps(TextWriter writer)
 		{
-			var isValid = componentName.Split('.').All(segment => _identifierRegex.IsMatch(segment));
-			if (!isValid)
+			//TODO: cache
```
dustinsoftware: Let's remove any TODOs before merging this in

src/React.Core/ReactComponent.cs
```diff
@@ -114,31 +101,61 @@ public ReactComponent(IReactEnvironment environment, IReactSiteConfiguration con
...
+						new StringWriter(new StringBuilder(512));
+				}
+				
 				try
 				{
-					var reactRenderCommand = renderServerOnly
-						? string.Format("ReactDOMServer.renderToStaticMarkup({0})", GetComponentInitialiser())
-						: string.Format("ReactDOMServer.renderToString({0})", GetComponentInitialiser());
-					html = _environment.Execute<string>(reactRenderCommand);
+					stringWriter.Write(renderServerOnly ? "ReactDOMServer.renderToStaticMarkup(" : "ReactDOMServer.renderToString(");
```
dustinsoftware: This doesn't look thread safe - multiple threads could call `Write` at the same time to this shared instance. Keeping `StringWriter` scoped to this method instead of using a static field would solve this.

DaniilSokolyuk: _sharedStringWriter is marked as ThreadStatic and this is unique for each thread. (corefx use similliar strategy, https://github.com/dotnet/corefx/blob/master/src/Common/src/System/IO/StringBuilderCache.cs) , I am reuse evertything for reduce allocations.

DaniilSokolyuk:
@dustinsoftware, yes you right, but previous ID generating is ugly, we are allocate GUID, after allocate byte array and 1-3 string...  my implementation used 1 string

src/React.Core/ReactComponent.cs
DaniilSokolyuk: we can return virtual but it does not make sense and and can cause errors
because all methods and extensions uses overload with textwriter

src/React.Core/ReactComponent.cs
DaniilSokolyuk: fixed

src/React.Core/ReactComponent.cs
dustinsoftware: Oh ok, did not know about ‘ThreadStatic’ :)

src/React.AspNet/HtmlHelperExtensions.cs
DaniilSokolyuk: yea, original tag builder has a similar code https://github.com/aspnet/Mvc/blob/feed0fea2c5280264ac000947b3fee542d7d9e6a/src/Microsoft.AspNetCore.Mvc.ViewFeatures/Rendering/TagBuilder.cs#L336

dustinsoftware:
> previous ID generating is ugly

There may be consuming code that depends on the current ID format, so this isn't a good reason to change it. We unfortunately don't know how consumers are using this project and want to avoid surprises between version updates.

Daniel15:
Thanks for reviewing this, @dustinsoftware!

src/React.Core/ReactComponent.cs
```diff
@@ -114,31 +101,61 @@ public ReactComponent(IReactEnvironment environment, IReactSiteConfiguration con
...
-			if (!renderContainerOnly)
+			if (_configuration.ComponentExistsChecks && !renderContainerOnly)
 			{
 				EnsureComponentExists();
 			}
 
 			var html = string.Empty;
 			if (!renderContainerOnly)
 			{
+				var stringWriter = _sharedStringWriter;
```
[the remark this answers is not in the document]
DaniilSokolyuk: 1. This variable is not unnecessary, because access to ThreadStatic variable is slow http://tips.x-tensive.com/2008/10/cost-of-threadstatic-attribute.html
2. StringBuilder is not implement TextWriter

src/React.Core/ReactComponent.cs
```diff
@@ -114,31 +101,61 @@ public ReactComponent(IReactEnvironment environment, IReactSiteConfiguration con
...
+				if (stringWriter != null)
+				{
+					stringWriter.GetStringBuilder().Clear();
+				}
+				else
+				{
+					_sharedStringWriter = 
+						stringWriter = 
+						new StringWriter(new StringBuilder(512));
```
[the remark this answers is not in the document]
DaniilSokolyuk: In part 2 i will remove all the StringWriter, StringBuilders and threadStatic and replace them with my own buffer based on ArrayPool, https://github.com/DaniilSokolyuk/React.NET/blob/c271afeb369875194a31c18bbbe6b76fac947d51/src/React.Core/ReactComponent.cs#L150

DaniilSokolyuk:
@damiangreen, we can announce this change or publish major version

dustinsoftware:
That’s not the point I was trying to make... I am hesitant to merge in changes for minor improvements that are technically breaking changes. 

The discussion on this PR is getting too long, please open a new one for each individual change (such as ones the bulleted list above) and we can discuss them one at a time :)

PR #528, opened 2018-04-08
Fast React ID Generator

DaniilSokolyuk:
Fast React ID Generator


src/React.Core/ReactIdGenerator.cs
```diff
@@ -0,0 +1,62 @@
...
+			if (chars == null)
+			{
+				_chars = chars = new char[19];
+				Array.Copy(reactPrefix, 0, chars, 0, reactPrefix.Length);
+			}
+
+			var id = Interlocked.Increment(ref _random);
+
+			//reactPrefix.Length == 6
+			chars[6] = _encode32Chars[(int)(id >> 60) & 31];
```
Daniel15: There's a lot of magic numbers in this code. It'd be good to pull these out as constants and add comments explaining what they're doing.

src/React.Core/ReactIdGenerator.cs
```diff
@@ -0,0 +1,62 @@
...
+
+using System;
+using System.Threading;
+
+namespace React
+{
+	/// <summary>
+	/// Extension methods relating to GUIDs.
+	/// </summary>
+	public static class ReactIdGenerator
```
Daniel15: Would be good to add an interface for this, and use the dependency injection framework to inject it into `ReactComponent`. That way, anyone can swap out the implementation if they want to (eg. if they want the old one back).

src/React.Core/ReactIdGenerator.cs
```diff
@@ -0,0 +1,62 @@
...
+		/// <summary>
+		/// Returns a short react identifier starts with "react_".
+		/// </summary>
+		/// <returns></returns>
+		public static string Generate()
+		{
+			var chars = _chars;
+			if (chars == null)
+			{
+				_chars = chars = new char[19];
```
Daniel15: Why 19?

src/React.Core/ReactIdGenerator.cs
```diff
@@ -0,0 +1,62 @@
...
+			chars[11] = _encode32Chars[(int)(id >> 35) & 31];
+			chars[12] = _encode32Chars[(int)(id >> 30) & 31];
+			chars[13] = _encode32Chars[(int)(id >> 25) & 31];
+			chars[14] = _encode32Chars[(int)(id >> 20) & 31];
+			chars[15] = _encode32Chars[(int)(id >> 15) & 31];
+			chars[16] = _encode32Chars[(int)(id >> 10) & 31];
+			chars[17] = _encode32Chars[(int)(id >> 5) & 31];
+			chars[18] = _encode32Chars[(int)id & 31];
+
+			return new string(chars, 0, 19);
```
Daniel15: Pull out `19` as a constant

src/React.Core/ReactIdGenerator.cs
Daniel15: (see `AssemblyRegistration.cs`, use `.AsSingleton()`)

DaniilSokolyuk: Nice idea, i can inject it to IReactEnvironment and create property IReactIdGenerator (like IBabel) and call it from constructor like "environment.ReactIdGenerator.Generate(...)"

src/React.Core/IReactIdGenerator.cs
```diff
@@ -7,18 +7,14 @@
...
  */
 
-using System;
-using Xunit;
-
-namespace React.Tests.Core
+namespace React
 {
-	public class GuidExtensionsTests
+	public interface IReactIdGenerator
```
dustinsoftware: Ugh, sorry. Missed this one. There is no XML comment for this type which is generating a warning at build time.

PR #530, opened 2018-04-08
Optionally skip component exists checks

DaniilSokolyuk:


Daniel15:
Are you doing this because the check is slow? If so, it might be better to create some JS wrappers around `ReactDOMServer.renderToStaticMarkup` and `ReactDOMServer.renderToString` that check for the component before calling the render function, and then replace the calls on lines 135-136 with calls to these new wrapper functions. That'd avoid one extra call from C# to JS, so should be a bit faster. You could add some functions to [shims.js](https://github.com/reactjs/React.NET/blob/master/src/React.Core/Resources/shims.js).

dustinsoftware:
This PR is getting stale and right now there isn't a strong need to merge it in. If there's something we're missing let me know and we can continue the discussion! 😄

PR #542, opened 2018-05-14
Updated the MVC tutorial for React 16

awayken:
I had issues that many others have had with getting the server rendering of this tutorial to work out of the box. I came to realize that it was a discrepancy in the version of React being used on the client versus in the NuGet package. NuGet was using the latest, version 16, but the client side code all referenced version 15. All sorts of confusion and chaos ensued.

This is a major change but it boils down to a shift over to ES6 classes instead of the older `React.createClass()` method. I tested as I went along, so the code should all work. I did not update screenshots. I think any reasonably seasoned .NET developer will skip over those details anyway. I am not a seasoned .NET developer, and I had no issues with that side of things.

I also added a note at the bottom explaining how server-side rendering can suffer from this version drift issue. Hopefully that will save people hours of debugging and frustration.

### DIFF
--- a/site/jekyll/getting-started/tutorial_aspnet4.md
+++ b/site/jekyll/getting-started/tutorial_aspnet4.md
@@ -94,7 +94,7 @@ React is all about modular, composable components. For our comment box example, 
   - CommentForm
 ```
 
-Let's build the `CommentBox` component, which is just displays a simple `<div>`. Add this code to `Tutorial.jsx`:
+Let's build the `CommentBox` component, which just displays a simple `<div>`. Add this code to `Tutorial.jsx`:
 
 ```javascript
 class CommentBox extends React.Component {
@@ -146,7 +146,7 @@ Its use is optional but we've found JSX syntax easier to use than plain JavaScri
 
 #### What's going on
 
-We are defining a new JavaScript class that extends from the React.Component class. In our class, we will define some properties and some methods to build from what React.Component already gives us. The most important of our methods is called `render` which returns a tree of React components that will eventually render to HTML.
+We are defining a new JavaScript class that extends from the React.Component class. In our class, we will define some properties and some methods to build from what React.Component already gives us. The most important of these methods is called `render` which returns a tree of React components that will eventually render to HTML.
 
 The `<div>` tags are not actual DOM nodes; they are instantiations of React `div` components. You can think of these as markers or pieces of data that React knows how to handle. React is **safe**. We are not generating HTML strings so XSS protection is the default.
 
@@ -472,7 +472,7 @@ This component is different from the prior components because it will have to re
 
 ### Reactive state
 
-So far, based on its props, each component has rendered itself once. `props` are immutable: they are passed from the parent and are "owned" by the parent. To implement interactions, we introduce a mutable **state** to the component. `this.state` is private to the component and can be changed by calling `this.setState()` and passing an object that represents changes in state. When the state updates, the component re-renders itself.
+So far, based on its props, each component has rendered itself once. `props` are immutable: they are passed from the parent and are "owned" by the parent. To implement interactions, we introduce mutable **state** to the component. `this.state` is private to the component and can be changed by calling `this.setState()` and passing an object that represents changes in state. When the state updates, the component re-renders itself.
 
 `render()` methods are written declaratively as functions of `this.props` and `this.state`. The framework guarantees the UI is always consistent with the inputs.
 
@@ -617,16 +617,16 @@ class CommentForm extends React.Component {
 
 #### Controlled components
 
-With the traditional DOM, `input` elements are rendered and the browser manages their state (its rendered value). As a result, the state of the actual DOM will differ from that of the component. This is not ideal as the state of the view will differ from that of the component. In React, components should always represent the state of the view and not only at the point of initialization.
+With the traditional DOM, `input` elements are rendered and the browser manages the state (its rendered value). As a result, the state of the actual DOM will differ from that of the component. This is not ideal as the state of the view will differ from that of the component. In React, components should always represent the state of the view and not only at the point of initialization.
 
 Hence, we will be using `this.state` to save the user's input as it is entered. We define an initial `state` with two properties `author` and `text` and set them to be empty strings. In our `<input>` elements, we set the `value` prop to reflect the `state` of the component and attach `onChange` handlers to them. These `<input>` elements with a `value` set are called controlled components. Read more about controlled components on the [Forms article](https://reactjs.org/docs/forms.html#controlled-components).
 
 ```javascript{2-13,16-28}
 class CommentForm extends React.Component {
   constructor(props) {
-	  super(props);
+    super(props);
     this.state = {author: '', text: ''};
-		this.handleAuthorChange = this.handleAuthorChange.bind(this);
+    this.handleAuthorChange = this.handleAuthorChange.bind(this);
     this.handleTextChange = this.handleTextChange.bind(this);
   }
   handleAuthorChange(e) {
@@ -670,11 +670,11 @@ Let's make the form interactive. When the user submits the form, we should clear
 ```javascript{7,15-24,27}
 class CommentForm extends React.Component {
   constructor(props) {
-	  super(props);
+    super(props);
     this.state = {author: '', text: ''};
-		this.handleAuthorChange = this.handleAuthorChange.bind(this);
+    this.handleAuthorChange = this.handleAuthorChange.bind(this);
     this.handleTextChange = this.handleTextChange.bind(this);
-		this.handleSubmit = this.handleSubmit.bind(this);
+    this.handleSubmit = this.handleSubmit.bind(this);
   }
   handleAuthorChange(e) {
     this.setState({author: e.target.value});
@@ -682,7 +682,7 @@ class CommentForm extends React.Component {
   handleTextChange(e) {
     this.setState({text: e.target.value});
   }
-	handleSubmit(e) {
+  handleSubmit(e) {
     e.preventDefault();
     const author = this.state.author.trim();
     const text = this.state.text.trim();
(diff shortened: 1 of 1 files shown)


site/jekyll/getting-started/tutorial_aspnet4.md
```diff
@@ -93,18 +94,19 @@ React is all about modular, composable components. For our comment box example,
   - CommentForm
 ```
 
-Let's build the `CommentBox` component, which is just a simple `<div>`. Add this code to `Tutorial.jsx`:
+Let's build the `CommentBox` component, which is just displays a simple `<div>`. Add this code to `Tutorial.jsx`:
```
dustinsoftware: > which is just displays a simple

This change doesn't make sense..

site/jekyll/getting-started/tutorial_aspnet4.md
```diff
@@ -465,24 +466,25 @@ ReactDOM.render(
...
 
 This component is different from the prior components because it will have to re-render itself. The component won't have any data until the request from the server comes back, at which point the component may need to render some new comments.
 
 ### Reactive state
 
-So far, based on its props, each component has rendered itself once. `props` are immutable: they are passed from the parent and are "owned" by the parent. To implement interactions, we introduce mutable **state** to the component. `this.state` is private to the component and can be changed by calling `this.setState()`. When the state updates, the component re-renders itself.
+So far, based on its props, each component has rendered itself once. `props` are immutable: they are passed from the parent and are "owned" by the parent. To implement interactions, we introduce a mutable **state** to the component. `this.state` is private to the component and can be changed by calling `this.setState()` and passing an object that represents changes in state. When the state updates, the component re-renders itself.
```
dustinsoftware: > we introduce a mutable **state**

This change seems to be unnecessary

site/jekyll/getting-started/tutorial_aspnet4.md
```diff
@@ -133,56 +135,58 @@ var CommentBox = React.createClass({displayName: 'CommentBox',
...
 );
 ```
 
-Its use is optional but we've found JSX syntax easier to use than plain JavaScript. Read more on the [JSX Syntax article](http://facebook.github.io/react/docs/jsx-in-depth.html).
+Its use is optional but we've found JSX syntax easier to use than plain JavaScript. Read more in React's ["JSX In Depth"](https://reactjs.org/docs/jsx-in-depth.html) article.
 
 #### What's going on
 
-We pass some methods in a JavaScript object to `React.createClass()` to create a new React component. The most important of these methods is called `render` which returns a tree of React components that will eventually render to HTML.
+We are defining a new JavaScript class that extends from the React.Component class. In our class, we will define some properties and some methods to build from what React.Component already gives us. The most important of our methods is called `render` which returns a tree of React components that will eventually render to HTML.
```
dustinsoftware: > The most important of our methods 

Unnecessary change, let's replace `our` with `these`

site/jekyll/getting-started/tutorial_aspnet4.md
```diff
@@ -604,27 +612,30 @@ var CommentForm = React.createClass({
...
     );
   }
-});
+}
 ```
 
 #### Controlled components
 
-With the traditional DOM, `input` elements are rendered and the browser manages the state (its rendered value). As a result, the state of the actual DOM will differ from that of the component. This is not ideal as the state of the view will differ from that of the component. In React, components should always represent the state of the view and not only at the point of initialization.
+With the traditional DOM, `input` elements are rendered and the browser manages their state (its rendered value). As a result, the state of the actual DOM will differ from that of the component. This is not ideal as the state of the view will differ from that of the component. In React, components should always represent the state of the view and not only at the point of initialization.
```
dustinsoftware: > browser manages the state

Unnecessary change, their => the

site/jekyll/getting-started/tutorial_aspnet4.md
```diff
@@ -604,27 +612,30 @@ var CommentForm = React.createClass({
...
-```javascript{2-10,14-25}
-var CommentForm = React.createClass({
-  getInitialState: function() {
-    return {author: '', text: ''};
-  },
-  handleAuthorChange: function(e) {
+```javascript{2-13,16-28}
+class CommentForm extends React.Component {
+  constructor(props) {
+	  super(props);
```
dustinsoftware: Indentation is off in this method

site/jekyll/getting-started/tutorial_aspnet4.md
```diff
@@ -643,41 +654,45 @@ var CommentForm = React.createClass({
...
-});
+}
 ```
 
 #### Events
 
 React attaches event handlers to components using a camelCase naming convention. We attach `onChange` handlers to the two `<input>` elements. Now, as the user enters text into the `<input>` fields, the attached `onChange` callbacks are fired and the `state` of the component is modified. Subsequently, the rendered value of the `input` element will be updated to reflect the current component `state`.
 
-(The astute reader may be surprised that these event handlers work as described, given that the method references are not explicitly bound to `this`. `React.createClass(...)` [automatically binds](/react/docs/interactivity-and-dynamic-uis.html#under-the-hood-autobinding-and-event-delegation) each method to its component instance, obviating the need for explicit binding.)
+You'll notice that we are explicitly binding our event handlers to `this` in the constructor. While older techniques, like `React.createClass(...)`, featured automatic binding, we are using ES6 classes to define our components. React components declared as ES6 classes don't automatically bind `this` to the instance, so we have to explicitly use `.bind(this)`.
```
dustinsoftware: You can also use class field syntax instead of binding methods in the constructor. Let's add a link to this page which describes alternate patterns further:

https://reactjs.org/docs/handling-events.html

site/jekyll/getting-started/tutorial_aspnet4.md
```diff
@@ -643,41 +654,45 @@ var CommentForm = React.createClass({
...
-  getInitialState: function() {
-    return {author: '', text: ''};
-  },
-  handleAuthorChange: function(e) {
+```javascript{7,15-24,27}
+class CommentForm extends React.Component {
+  constructor(props) {
+	  super(props);
+    this.state = {author: '', text: ''};
+		this.handleAuthorChange = this.handleAuthorChange.bind(this);
```
dustinsoftware: Indentation

site/jekyll/getting-started/tutorial_aspnet4.md
```diff
@@ -696,41 +711,41 @@ var CommentForm = React.createClass({
...
-
-Call `preventDefault()` on the event to prevent the browser's default action of submitting the form.
+We attach an `onSubmit` handler to the form that clears the form fields when the form is submitted with valid input. We call `preventDefault()` on the event to prevent the browser's default action of submitting the form.
 
 #### Callbacks as props
 
 When a user submits a comment, we will need to refresh the list of comments to include the new one. It makes sense to do all of this logic in `CommentBox` since `CommentBox` owns the state that represents the list of comments.
 
-We need to pass data from the child component back up to its parent. We do this in our parent's `render` method by passing a new callback (`handleCommentSubmit`) into the child, binding it to the child's `onCommentSubmit` event. Whenever the event is triggered, the callback will be invoked:
+We need to pass data from the child component back up to its parent. We do this in our parent's `render` method by passing a new callback (`handleCommentSubmit`) into the child and binding it to the child's `onCommentSubmit` event. Whenever the event is triggered, the callback will be invoked:
```
dustinsoftware: > into the child and binding

Unnecessary change

site/jekyll/getting-started/tutorial_aspnet4.md
```diff
@@ -696,41 +711,41 @@ var CommentForm = React.createClass({
...
-```javascript{11-13,26}
-var CommentBox = React.createClass({
-  loadCommentsFromServer: function() {
-    var xhr = new XMLHttpRequest();
+```javascript{5,16-18,28}
+class CommentBox extends React.Component {
+  constructor(props) {
+    super(props);
+    this.state = { data: [] };
+		this.handleCommentSubmit = this.handleCommentSubmit.bind(this);
```
dustinsoftware: Indentation

site/jekyll/getting-started/tutorial_aspnet4.md
```diff
@@ -739,33 +754,37 @@ var CommentBox = React.createClass({
...
-  getInitialState: function() {
-    return {author: '', text: ''};
-  },
-  handleAuthorChange: function(e) {
+```javascript{22}
+class CommentForm extends React.Component {
+  constructor(props) {
+	  super(props);
+    this.state = {author: '', text: ''};
+		this.handleAuthorChange = this.handleAuthorChange.bind(this);
```
dustinsoftware: Indentation

site/jekyll/getting-started/tutorial_aspnet4.md
```diff
@@ -961,44 +981,45 @@ If you go to this URL in your browser, you should notice that the code has been
...
+    };
     xhr.send();
-  },
-  handleCommentSubmit: function(comment) {
-    var comments = this.state.data;
-    var newComments = comments.concat([comment]);
+  }
+	handleCommentSubmit(comment) {
+    const comments = this.state.data;
+    comment.Id = comments.length + 1;
```
dustinsoftware: This looks like a bug fix, was it intentional?

site/jekyll/getting-started/tutorial_aspnet4.md
```diff
@@ -1071,14 +1104,14 @@ namespace ReactDemo
...
-				.AddScript("~/js/remarkable.min.js")
+				.AddScript("~/Scripts/remarkable.min.js")
 				.AddScript("~/Scripts/Tutorial.jsx");
 		}
 	}
 }
 ```
 
-Note that we need a copy of Remarkable in order to load it for server-side rendering. In a production app you'd probably use Bower or npm for this, but for our tutorial you can just [download the file from CDNJS](https://cdnjs.cloudflare.com/ajax/libs/remarkable/1.7.1/remarkable.min.js) and save it into `~/js`.
+Note that we need a copy of Remarkable in order to load it for server-side rendering. In a production app you'd probably use Bower or npm for this, but for our tutorial you can just [download the file from CDNJS](https://cdnjs.cloudflare.com/ajax/libs/remarkable/1.7.1/remarkable.min.js) and save it into `~/Scripts`.
```
dustinsoftware: Leaving remarkable in the `js` folder should prevent it from being unnecessarily compiled through babel.

site/jekyll/getting-started/tutorial_aspnet4.md
awayken: Yeah. React DevTools was yelling at me about using a key, and I couldn't figure out what was wrong because you were already using a key. Then I saw the property difference and realized what was happening.

site/jekyll/getting-started/tutorial_aspnet4.md
awayken: Okay. I can put it back.

site/jekyll/getting-started/tutorial_aspnet4.md
awayken: I think this would still be good to include. I might put out another PR discussing this alternative. I didn't get it into this one before you merged it, but I can see this eventually being the preferred syntax once it stops being experimental.

awayken:
@dustinsoftware Happy to help! I've just started using React.NET for a project I'm working on, so I'll definitely take a look.

dustinsoftware:
I just noticed after merging this that the .NET Core docs didn't get updated at all... will need to port these changes over.

Daniel15:
Yeah it's unfortunate that there's a division between the .NET Core and 'classic' .NET Framework and they can get out of sync. Maybe we can pull out shared stuff into separate .md files and include them in both pages.

PR #544, opened 2018-05-17
Add detail about alternate ways of handling events

awayken:


Daniel15:
Thanks! It'd be good to add this to the ASP.NET Core tutorial page too.

PR #548, opened 2018-06-10
Update docs

dustinsoftware:
This PR makes both the .NET Core and .NET Framework tutorials consistent, plus improves a few other things.


dustinsoftware:
Ooh, netlify support in PR's? 🤤

Daniel15:
@dustinsoftware Yeah I'm playing around with Netlify. An issue is that we have both the code and the site in the same repo, but I only want Netlify to post comments on PRs when they touch files in the site. Not sure if that's possible at the moment.

PR #558, opened 2018-06-29
Fix crashes related to loading React 16.4.1

### LINKED ISSUE #555: Crash when using React 16.4.1
While investigating #554, I discovered that React.NET totally breaks when using `react` and `react-dom` 16.4.1, but 16.4.0 seems to be unaffected, so users can pin to that version until we ship a fix.

Unfortunately the crash is obscured by a check that looks for React on the global scope. To see the real exception, patch `ReactNET_initReact` to always return `true` in `shims.js`. I'll open up a separate issue to investigate whether we still need to have that method around (we'd just crash at runtime if one of the globals was missing, right?)

Anyhow. The real crash is `ReactScriptLoadException: Error while loading "~/components-bundle.generated.js": ReferenceError: 'setTimeout' is not defined`

Which corresponds to `SelectEventPlugin`:
```js
// We capture a local reference to any global, in case it gets polyfilled after
// this module is initially evaluated.
// We want to be using a consistent implementation.
var localDate = Date;
var localSetTimeout = setTimeout; // crash!
var localClearTimeout = clearTimeout;
```

The regression appears to be introduced by this change: https://github.com/facebook/react/pull/12931/files#diff-bbebc3357e1fb99ab13ad796e04b69a6L47

I'll open an issue tomorrow and investigate a fix. cc @Daniel15

dustinsoftware:
Fixes #555

### DIFF
--- a/src/React.Core/JavaScriptEngineFactory.cs
+++ b/src/React.Core/JavaScriptEngineFactory.cs
@@ -135,7 +135,7 @@ namespace React
 			}
 
 			LoadUserScripts(engine);
-			if (!_config.LoadReact)
+			if (!_config.LoadReact && _scriptLoadException == null)
 			{
 				// We expect to user to have loaded their own version of React in the scripts that
 				// were loaded above, let's ensure that's the case. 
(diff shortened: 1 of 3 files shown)


Daniel15:
Thanks!

PR #568, opened 2018-07-10
Bumped to latest minor React versions in React.Core package.json

### LINKED ISSUE #566: Please update the built-in React to 16.3 or higher
Following #564: ~~you're~~ React.Core.3.3.0 is on 16.0. https://github.com/reactjs/React.NET/blob/6f6ae01135f8abd709d80a2b470bcd2dde1f285d/src/React.Core/package.json#L15

React 16.3 added `React.createRef` _(which I happen to rely on)_ and generally newer versions have bug fixes, API improvements, etc.

JoshuaKGoldberg:
I held off breaking API changes such as moving Webpack from 1 to 4 (!) out of laziness.

Fixes #566.

Edit: Err, `Map is not defined` build errors. Reducing scope to just React.


JoshuaKGoldberg:
```
ReferenceError: Map is not defined [C:\projects\react-net\src\React.Sample.Mvc4\React.Sample.Mvc4.csproj]
C:\projects\react-net\src\React.Sample.Mvc4\TransformBabel.proj(5,3): error MSB4018: at React.Core.Resources.react.generated.min.js:4:69
```

I don't see why this is happening... no plans to investigate on my end, but I'll leave the PR open for visibility. Sorry :)

dustinsoftware:
We should update the webpack samples before merging this.

dustinsoftware:
Updating to React 16.4 breaks at least one JS engine that is loaded by default (VroomJsEngine). Need to audit the samples and fix any failures before merging.

Taritsyn:
@dustinsoftware The VroomJs library uses a very old version of V8. Version 3.17.16.2 was released more than 5 years ago (April 11, 2013).

dustinsoftware:
I think we should just retire the engine specific logic in react.core,
there is a PR that did exactly that which just needs a bit of polish before
we can merge it.

Taritsyn:
@dustinsoftware If you're talking about [this PR](https://github.com/reactjs/React.NET/pull/501), then I don't like the following block of source code:

```csharp
///<summary>
/// Gets or sets the name of the default JsEngineSwitcher engine.
/// </summary>
public string DefaultEngineName
{
	get
	{
		return JsEngineSwitcher.Instance.DefaultEngineName;
	}
	set
	{
		JsEngineSwitcher.Instance.DefaultEngineName = value;
	}
}
```

It is not safe to change the `DefaultEngineName` property of the JavaScript Engine Switcher, which other libraries can use, from the ReactJS.NET. I would name this property `EngineName` and set the default value to an empty string. In case the user did not set a value to this property, I would try to take the value from the `JsEngineSwitcher.Instance.DefaultEngineName`. If it is also empty, then I would throw an exception.

dustinsoftware:
Fixed in fe72ea9ca18793fea75282f0514e969384853d5e

PR #571, opened 2018-07-11
Returned a list of missing globals in shims.js

### LINKED ISSUE #569: State which of [React, ReactDOM, and ReactDOMServer] aren't loaded correctly in shims.js
Following #566 by including manually including React .js files with `AddScript`, I've included all three `.js` files from React's UMD build and am having a jolly time trying to get around:

```
React.Exceptions.BabelException Exception Message:
    In file "Content/webclient/js/dependencies/react.development.js":
        React has not been loaded correctly. Please expose your version of React as global variables named 'React', 'ReactDOM' and 'ReactDOMServer', or enable the 'LoadReact' configuration option to use the built-in version of React. 
```

Could `shims.js` explain _which_ React globals are missing? That'd be very helpful!

JoshuaKGoldberg:
`EnsureReactLoaded` uses this to print out which ones are missing.

Fixes #569.

### DIFF
--- a/src/React.Core/Resources/shims.js
+++ b/src/React.Core/Resources/shims.js
@@ -1,4 +1,4 @@
-﻿/*
+/*
  *  Copyright (c) 2014-Present, Facebook, Inc.
  *  All rights reserved.
  *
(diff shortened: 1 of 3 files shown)


src/React.Core/Resources/shims.js
```diff
@@ -46,26 +46,37 @@ if (!Object.freeze) {
...
-		return true;
+	if (typeof ReactDOMServer === 'undefined') {
+		if (global.ReactDOMServer) {
+			ReactDOMServer = global.ReactDOMServer;
+		}
+		else {
+			missing.push('ReactDOMServer');
+		}
 	}
-	// :'(
```
JoshuaKGoldberg: I was very sad removing this... but couldn't think of a way to keep it in. :'(

dustinsoftware:
Argh, this should have been a squash. Oh well. Thanks :)

JoshuaKGoldberg:
Whoohoo, thanks! ✨

@dustinsoftware is there a schedule/plan for releasing this? I'd love to have a patch version if you don't mind, just to simplify local development.

PR #574, opened 2018-07-23
Fix for the `EnsureReactLoaded` method of `React.JavaScriptEngineFactory` class

Taritsyn:
Hello!

The following [line of code](https://github.com/reactjs/React.NET/blob/83785cf9ea6cea01341ca023a32b4086d5a14cfb/src/React.Core/JavaScriptEngineFactory.cs#L188):

```csharp
var result = engine.CallFunction<string[]>("ReactNET_initReact");
```

Will always cause an error:

```
System.ArgumentException: The type of return value `System.String[]` is not supported.
```

This error occurs, because the JavaScript Engine Switcher only supports primitive types. Here is a quote from the documentation:

> The supported .NET types are as follows:
> 
>  * `JavaScriptEngineSwitcher.Core.Undefined`
>  * `System.Boolean`
>  * `System.Int32`
>  * `System.Double`
>  * `System.String`

This PR contains the simplest fix for this error.

### DIFF
--- a/src/React.Core/JavaScriptEngineFactory.cs
+++ b/src/React.Core/JavaScriptEngineFactory.cs
@@ -185,11 +185,13 @@ namespace React
 		/// <param name="engine">Engine to check</param>
 		private static void EnsureReactLoaded(IJsEngine engine)
 		{
-			var result = engine.CallFunction<string[]>("ReactNET_initReact");
-			if (result.Length != 0)
+			var globalsString = engine.CallFunction<string>("ReactNET_initReact");
+			string[] globals = globalsString.Split(new[] { ',' }, StringSplitOptions.RemoveEmptyEntries);
+
+			if (globals.Length != 0)
 			{
 				throw new ReactNotInitialisedException(
-					$"React has not been loaded correctly: missing ({string.Join(", ", result)})." +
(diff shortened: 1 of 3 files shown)


dustinsoftware:
Thanks @Taritsyn. I tested this before merging, not sure how I did not see this error. Will do some digging later to see what happened.

Daniel15:
Probably another reason why we need more integration tests - I suspect this functionality is mocked out in all the automated tests. I'll push a beta version with this fix later today.

Taritsyn:
@dustinsoftware This error really passes all the tests. I was able to reproduce it only using the following setting:

```csharp
	…
	.SetLoadReact(false)
	…
	;
```

dustinsoftware:
I'll see if we can get a simple test added that doesn't require Browserstack or Selenium for the server-rendering pipeline. The `React.Sample.CoreMvc` project runs fine, but the `React.Sample.Router.CoreMvc` sample does not (which depends on a bundled copy of React).

I confirmed that @Taritsyn's fix works with the router sample.

Daniel15:
@dustinsoftware It should be possible to test the server-side rendering as long as a JS engine works on AppVeyor (which it should). We just need to have some tests that do not mock out the JS engine.

PR #578, opened 2018-08-05
Update tests

dustinsoftware:
This catches regressions like the one introduced in #571


dustinsoftware:
@Daniel15 got some simple server-render tests added. Any other integration test cases you think we should test for?

Daniel15:
This looks great! Thanks @dustinsoftware!

I wonder if we should have tests with multiple engines (eg. a test with V8, a test with ChakraCore)?

PR #580, opened 2018-08-12
Fix JS engine reloading in .NET Core

### LINKED ISSUE #554: Server side rendered components require server restart to take effect
Using ASP.NET Core 2.1
Webpack 4
Typescript 2.9

dustinsoftware:
Fixes #554 

It looks like this has never worked correctly with .NET Core...

JSPool looks for an exact match when a file changed event is raised, but we weren't first transforming the forwardslashes into backslashes.

asdfasdfasdf

### DIFF
--- a/src/React.AspNet.Middleware/AspNetFileSystem.cs
+++ b/src/React.AspNet.Middleware/AspNetFileSystem.cs
@@ -1,9 +1,9 @@
-﻿/*
+/*
  *  Copyright (c) 2015, Facebook, Inc.
  *  All rights reserved.
  *
  *  This source code is licensed under the BSD-style license found in the
- *  LICENSE file in the root directory of this source tree. An additional grant 
+ *  LICENSE file in the root directory of this source tree. An additional grant
  *  of patent rights can be found in the PATENTS file in the same directory.
  */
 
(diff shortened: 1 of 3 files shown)


src/React.AspNet.Middleware/AspNetFileSystem.cs
```diff
@@ -36,12 +36,12 @@ public AspNetFileSystem(IHostingEnvironment hostingEnv)
...
-            {
-                return relativePath;
-            }
-            relativePath = relativePath.TrimStart('~').TrimStart('/');
-            return Path.Combine(_hostingEnv.WebRootPath, relativePath);
+			if (relativePath.StartsWith(_hostingEnv.WebRootPath))
+			{
+				return relativePath;
+			}
+			relativePath = relativePath.TrimStart('~').TrimStart('/').Replace("/", "\\");
```
Daniel15: Shouldn't `Path.Combine` be normalizing the path already?

dustinsoftware: Unfortunately it doesn't fix slashes :( :(

dustinsoftware: Oh! Path.GetFullPath does convert slashes, so that's a more elegant fix.

dustinsoftware:
Tested on both OS X and Windows. @Daniel15 can you cut a release with this fix?

Daniel15:
Sure, I can do that once I'm home tonight.

Daniel15:
Actually I won't get time tonight but I can do it tomorrow.

Daniel15:
Preparing this for release now. Just cherry-picking into the stable branch and verifying that everything is OK.

I wonder if JSPool should normalize the path itself, too. What do you think @dustinsoftware?

dustinsoftware:
That thought occurred to me as well. It seems like JSPool should either
assert that the path is well formatted (no forward slashes on windows or
backslashes on non-Windows), or attempt to normalize it (which could hide
bugs). If the former is straight forward to do then that would probably be
better.

Issue #584, opened 2018-08-18
Can react-jss can be used with reactjs.net ?


DudeM29:
I am trying to use react jss in my server side rendring app but it seem the css is not being attached to the page

dustinsoftware:
Do the styles appear after the client render completes?

Saturate:
Yes, take a look at #602

Issue #487, opened 2017-12-08
Engines are not returned to pool "per request" but rather "per Html.React method invocation"


ryedin:
see: https://github.com/reactjs/React.NET/blob/master/src/React.AspNet/HtmlHelperExtensions.cs#L84

I'm wondering why the engine is being returned to the pool after each call to `Html.React`, when in all of the comments/docs you indicate that it should be "per request" (and I believe a true "per request" model would be much better overall performance).

Performance concerns aside, the way this is working right now also prevents using things like reactive mobx stores on the server across calls to `Html.React` within a given request context. We want to be able to disconnect components from the data that drives them and use dependency injection (on the js side) to connect them to the data. This prevents us from having to explicitly push ALL props in at the top level (the so-called page-level component). This works fine for the client-side rendering, as expected, because there is only a single js execution context. We had assumed that there would also be a single js execution context on the server (per request), because that would make a lot of sense. However, we soon realized that our injected data (injected via subsequent calls to `Html.React` in various partials) is _never_ made available, and is thus not actually being utilized for SSR.

Are there other reasons that I am not able to see that are preventing making the engine be truly a singleton across the _entire_ request, rather than simply per call to `Html.React`?

ryedin:
Got this to work as expected by disabling engine reuse in the config, but IMO, we should be able to configure pooling AND allow the same engine instance to persist for the entire request... and that should be the default behavior.

But again, please let me know if there is an explicit reason why you feel the engine(s) should be recycled per call to `Html.React` even if you're still in the same request.

dustinsoftware:
This behavior changed in #270. 

There is discussion there - it was possible for engines not to be returned to the pool in some cases, which caused a fatal crash when all the engines are gone, so that change avoids the issue entirely. 

I suppose we could make this configurable but I’m not sure there’s much value encouraging this aside from your specific use case for sharing engine scope across multiple requests, as a render call now counts on side-effects from previous render calls.

ryedin:
> use case for sharing engine scope across multiple requests

You keep using the word `request` when you really mean `method calls within the context of a single request`. This isn't about sharing state `across requests`. It's about using the semantics of a `single` request to build up state as the HTML is created that will be sent back to the client. This is not uncommon, nor without value, nor should it be "not encouraged".

[...]

PR #589, opened 2018-08-25
Add server-render support for CSS-in-JS

### LINKED ISSUE #538: Add support for styled-components
[Styled components](https://github.com/styled-components/styled-components) is a fairly popular library for including styles inside components. It generates a set of styles at runtime and appends them to the head tag, solving many problems with CSS at scale.

This issue is for tracking support for server-rendering styles with React.NET. Server-rendering for node is already supported and documented [here](https://www.styled-components.com/docs/advanced#server-side-rendering).

Goals:
- Styles are rendered to the head element
- Implementation isn't tied directly to styled-components
- No additional package is needed to consume this logic

I built a prototype and pushed my work [here](https://github.com/dustinsoftware/React.NET/tree/styled-components). It needs more polish before I open a real pull request, but I wanted to discuss the strategy here before getting too far along.

The proposed implementation isn't tied to `styled-components`. Instead, a wrapper function is provided to the render call, which will be invoked instead of `renderToString` from `react-dom/server`.

```cshtml
@Html.React(componentName, props, wrapRenderFunction: "renderStyles");`
```

The consumer defines the function on the global scope and exposes it via a script (via webpack most likely)

```js
import { ServerStyleSheet } from 'styled-components';
import { renderToString } from 'react-dom/server';

export default function renderWithStyles(renderFunc) {
	const sheet = new ServerStyleSheet();
	const html = renderToString(sheet.collectStyles(renderFunc));
	return JSON.stringify({ html, styles: sheet.getStyleTags() });
}
```

Finally, the styles are rendered from the razor layout:

```cshtml
<html>
<head>
	@RenderSection("Head", required: false)
	@Html.ReactGetStyles()
</head>
<body>
...
</body>
</html>
```

dustinsoftware:
Fixes #538 #584 

This is ready for API review. Once we're OK with the general direction, I'll put some time into polishing it up. It is not tied to any particular CSS-in-JS implementation. Rather, it provides some hooks that a caller can use to register a server stylesheet object to render into, and then fetch data out of that stylesheet.

You could also use this API for some generic data set up, such as #487. This is mostly intended as an escape hatch to run arbitrary Javascript on the same engine that is about to render the component.

This needs a bit more testing with other CSS-in-JS libraries (such as aphrodite). I only tested it with styled-components.

We could make this API a bit less generic, and register the functions for the lifetime of the app instead of passing them in to the Html.Render helper method. The downside is there needs to be a way to access the current `ViewBag` for the request so that the stylesheet can be written on the rendered layout.


src/React.AspNet/HtmlHelperExtensions.cs
```diff
@@ -63,7 +63,10 @@ private static IReactEnvironment Environment
 			bool clientOnly = false,
 			bool serverOnly = false,
 			string containerClass = null,
-			Action<Exception, string, string> exceptionHandler = null
+			Action<Exception, string, string> exceptionHandler = null,
+			Action<Func<string, string>> preRender = null,
+			Func<string, string> transformRender = null,
+			Action<Func<string, string>> postRender = null
```
Daniel15: This function signature is getting a bit long... But I guess it's OK since they're all optional.

src/React.Core/IReactComponent.cs
```diff
@@ -55,7 +55,7 @@ public interface IReactComponent
 		/// <param name="renderServerOnly">Only renders the common HTML mark up and not any React specific data attributes. Used for server-side only rendering.</param>
 		/// <param name="exceptionHandler">A custom exception handler that will be called if a component throws during a render. Args: (Exception ex, string componentName, string containerId)</param>
 		/// <returns>HTML</returns>
-		string RenderHtml(bool renderContainerOnly = false, bool renderServerOnly = false, Action<Exception, string, string> exceptionHandler = null);
+		string RenderHtml(bool renderContainerOnly = false, bool renderServerOnly = false, Action<Exception, string, string> exceptionHandler = null, Action<Func<string, string>> preRender = null, Func<string, string> transformRender = null, Action<Func<string, string>> postRender = null);
```
Daniel15: Split this onto multiple lines like in `HtmlHelperExtensions`?

src/React.Sample.Router.CoreMvc/Views/Home/Index.cshtml
```diff
@@ -1,15 +1,10 @@
...
-	<title>React Router Sample</title>
-	<meta charset="utf-8" />
-</head>
-<body>
-	@Html.ReactRouterWithContext("RootComponent", new { })
+@{
+    Layout = "_Layout";
+}
+
+@Html.ReactRouter("RootComponent", new { }, preRender: x => x("var serverStyleSheet = new Styled.ServerStyleSheet();"), transformRender: (component) => $"serverStyleSheet.collectStyles({component})", postRender: x => ViewBag.ServerStyles = x("serverStyleSheet.getStyleTags()"))
```
Daniel15: Hmm, I wonder if it's worth having something inbuilt for styled-components specifically, for example maybe an API like this? 
```
Html.ReactRouter("RootComponent", new { })
  .WithStyledComponents()
```

awayken:
I'd be interested in seeing this feature added. The first thing I did on my latest React.net project was to explore CSS-in-JS options and soon ran into issues (and then #538 's explanation).

I'll see if I can find some time to test with my preferred styling library.

dustinsoftware:
I haven’t forgotten about this, just have spent my recent evenings and
weekends doing home remodel work :) I should have time this weekend

src/React.AspNet/HtmlHelperExtensions.cs
dustinsoftware: I'll see about moving these into their own type, since they are all optional render settings.

dustinsoftware: I ended up moving most of these into a separate interface

src/React.Sample.Router.CoreMvc/Views/Home/Index.cshtml
dustinsoftware: Implemented as a chainable helper class [here](https://github.com/reactjs/React.NET/pull/589/files#diff-2a8446f860d13d661d944a659e7516c1R5)

dustinsoftware:
I restructured how the render helper functions got passed around and moved them into an interface and some helper classes. This PR demos using both styled components and react router in the same project with these new helpers, and allows new ones to be added either inside this library or by third parties easily.

dustinsoftware:
This has been in production for about a week on one of my team's projects and has worked well. I'm going to push up some method documentation, add some tests, and then merge it. I'd like to get some docs pages added as well detailing how to use this feature but that might come after this PR is merged.

dustinsoftware:
Opening a separate PR to update the samples

Saturate:
Would be nice with en example on how to do SSR with https://github.com/cssinjs/react-jss I'm having a hard time wrapping my head around this.

PR #592, opened 2018-08-29
Add ability to pre-compile scripts

Taritsyn:
Since version 3.X Beta the following modules of the JavaScript Engine Switcher support a script pre-compilation: JavaScriptEngineSwitcher.Msie (only in the Chakra JsRT modes), JavaScriptEngineSwitcher.V8, 
JavaScriptEngineSwitcher.Jurassic, JavaScriptEngineSwitcher.Jint and JavaScriptEngineSwitcher.ChakraCore. This feature allows you to once pre-compile the script and many times use it for initialization of engine instances. Due the fact that the pre-compiled script does not require parsing, you can significantly speed up the initialization of engines.

This PR adds this feature to the ReactJS.NET library. Pre-compilation is enabled by using the `AllowJavaScriptPrecompilation` property (default `false`). Also, pre-compilation requires the using of the caching provider is different from the `NullCache` type.


dustinsoftware:
Cool! I’ll try to look at this in the next few days

src/React.Core/JavaScriptEngineFactory.cs
```diff
@@ -138,6 +145,25 @@ protected virtual void InitialiseEngine(IJsEngine engine)
 			}
 		}
 
+		/// <summary>
+		/// Loads a code from embedded JavaScript resource into the engine.
```
dustinsoftware: Wording: "Loads a code" -> "Loads code"

src/React.Core/JavaScriptEngineUtils.cs
```diff
@@ -61,6 +77,151 @@ public static bool EnvironmentSupportsClearScript()
...
+			if (!engine.SupportsScriptPrecompilation)
+			{
+				Trace.WriteLine(string.Format("The {0} version {1} does not support the script pre-compilation.",
+					engine.Name, engine.Version));
+				return false;
+			}
+
+			if (cache is NullCache)
+			{
+				Trace.WriteLine("Usage of script pre-compilation without caching does not make sense.");
```
dustinsoftware: Should this be an error instead? If someone uses `NullCache` then and tries to use this logic, then they didn't configure it correctly and might miss this trace message...

dustinsoftware:
The benchmarks run fine on OS X but crash on Windows only when script precompilation is enabled. Going to investigate this further..

Taritsyn:
> The benchmarks run fine on OS X but crash on Windows only when script precompilation is enabled.

In ChakraCore .NET wrapper for Windows and Unix uses different code:

```csharp
if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
{
	…
}
else
{
	…
}
```

dustinsoftware:
This is concerning - it looks like within benchmarkdotnet, enabling script precompilation causes the entire process to die. Not sure why this would be limited to chakracore...

Taritsyn:
I think that we can do without the `InProcessToolchain`. Updating the BenchmarkDotNet to version 0.11.1 fixes the previously described errors.

However, if we assign to `ReuseJavaScriptEngines` property value equal to `true`, the errors return back.

Taritsyn:
To fix this error:

```
// * Warnings *
Environment
  Summary -> Detected error exit code from one of the benchmarks. It might be caused by following antivirus software:
        - Windows Defender (windowsdefender://)
        - Norton Security (C:\Program Files\Norton Security\Engine\22.15.0.88\WSCStub.exe)
Use InProcessToolchain to avoid new process creation.
```

I had to completely [abandon usage of the old ChakraCore API for Windows](https://github.com/Taritsyn/JavaScriptEngineSwitcher/commit/8b1fe692d148af3a14bb2f8df1c9fd8f38e2adfd#diff-c5c6c5b4817fde2ef25e61a1d9be4115) (Internet Explorer-like API).

src/React.Core/JavaScriptEngineUtils.cs
```diff
@@ -61,6 +77,151 @@ public static bool EnvironmentSupportsClearScript()
...
+				return false;
+			}
+
+			return true;
+		}
+
```
Daniel15: It'd be good to put all this code into a separate file (eg. `JavaScriptPrecompilation.cs`) to keep things organized.

src/React.Core/ReactSiteConfiguration.cs
```diff
@@ -216,17 +216,19 @@ public IReactSiteConfiguration SetMaxUsagesPerEngine(int? maxUsagesPerEngine)
 		}
 
 		/// <summary>
-		/// Gets or sets whether the MSIE engine should be used if V8 is unavailable.
+		/// Gets or sets whether to allow the JavaScript pre-compilation (accelerates the
+		/// initialization of JavaScript engines).
 		/// </summary>
-		public bool AllowMsieEngine { get; set; }
+		public bool AllowJavaScriptPrecompilation { get; set; }
```
Daniel15: This is a breaking change (removing a field from the public API) so we need to ensure we do a major version bump for this. However, I think we are already bumping the major version due to using the latest beta JavaScriptEngineSwitcher, so it should be fine.

tests/React.Tests.Common/MemoryFileCacheCore.cs
```diff
@@ -0,0 +1,81 @@
...
+			{
+				SlidingExpiration = slidingExpiration,
+			};
+
+			_cache.Set(key, data, options);
+		}
+	}
+}
+
+#endif
```
Daniel15: Is this code specific to tests?

dustinsoftware:
Yeah, the memory file cache is for tests only. I can add a comment to that
file to make that clearer.

src/React.Core/ReactSiteConfiguration.cs
Taritsyn: `AllowMsieEngine` property was removed from the `IReactSiteConfiguration` interface in the [“Remove auto-registered JS engines from React.Core”](https://github.com/reactjs/React.NET/pull/582) PR. I just removed the unused code.

Taritsyn: I plan to release a new version of the JavaScript Engine Switcher (Release Candidate 1) in next few days.

src/React.Core/JavaScriptEngineUtils.cs
Taritsyn: I think that for the first check it is necessary to generate an exception too.

dustinsoftware:
🚀 Works great for .JS files (`AddScriptWithoutTransform`), however I discovered that it does not work at all for files that Babel has to process (`AddScript`). It would be very cool if that was supported, but if that's not straight forward to do before merge, we'll need to add a comment indicating this behavior.

Benchmarks for `AddScriptWithoutTransform`:

``` ini

BenchmarkDotNet=v0.10.14, OS=macOS High Sierra 10.13.6 (17G65) [Darwin 17.7.0]
Intel Core i7-4980HQ CPU 2.80GHz (Haswell), 1 CPU, 8 logical and 4 physical cores
.NET Core SDK=2.0.0
  [Host]     : .NET Core 2.0.0 (CoreCLR 4.6.00001.0, CoreFX 4.6.25519.03), 64bit RyuJIT
  DefaultJob : .NET Core 2.0.0 (CoreCLR 4.6.00001.0, CoreFX 4.6.25519.03), 64bit RyuJIT


```
|                      Method | withPrecompilation |     Mean |     Error |    StdDev |   Gen 0 | Allocated |
|---------------------------- |------------------- |---------:|----------:|----------:|--------:|----------:|
| **Environment_CreateComponent** |              **False** | **48.49 ms** | **0.3031 ms** | **0.2687 ms** | **62.5000** |   **1.54 MB** |
| **Environment_CreateComponent** |               **True** | **41.63 ms** | **0.2408 ms** | **0.2134 ms** | **62.5000** |   **1.27 MB** |

dustinsoftware:
Also is there any reason *not* to enable this by default?

Taritsyn:
Hello, Dustin!

> …, however I discovered that it does not work at all for files that Babel has to process (`AddScript`). It would be very cool if that was supported, but if that's not straight forward to do before merge, we'll need to add a comment indicating this behavior.

For several days I'm working on a new version of the JavaScriptEngineSwitcher.ChakraCore module. If run the latest version of benchmarks together with new version of the JavaScriptEngineSwitcher.ChakraCore module ([891da69](https://github.com/Taritsyn/JavaScriptEngineSwitcher/commit/891da6911a5a36d56a0fd4f65dc8f1dbd1a15347)) on Windows, can get the following results:

``` ini
BenchmarkDotNet=v0.10.14, OS=Windows 10.0.17134
Intel Core i7 CPU 870 2.93GHz (Nehalem), 1 CPU, 8 logical and 4 physical cores
Frequency=2857588 Hz, Resolution=349.9455 ns, Timer=TSC
.NET Core SDK=2.0.0
  [Host]     : .NET Core 2.0.7 (CoreCLR 4.6.26328.01, CoreFX 4.6.26403.03), 32bit RyuJIT
  DefaultJob : .NET Core 2.0.7 (CoreCLR 4.6.26328.01, CoreFX 4.6.26403.03), 32bit RyuJIT
```

**ComponentRenderWithBabelBenchmarks**

|                      Method | withPrecompilation |     Mean |     Error |    StdDev |   Gen 0 | Allocated |
|---------------------------- |------------------- |---------:|----------:|----------:|--------:|----------:|
| **Environment_CreateComponent** |              **False** | **58.40 ms** | **1.0167 ms** | **0.9510 ms** | **62.5000** |   **2.86 MB** |
| **Environment_CreateComponent** |               **True** | **57.38 ms** | **0.8534 ms** | **0.7983 ms** | **62.5000** |   **2.86 MB** |

**ComponentRenderWithoutBabelBenchmarks**

|                      Method | withPrecompilation |     Mean |    Error |    StdDev |    Gen 0 | Allocated |
|---------------------------- |------------------- |---------:|---------:|----------:|---------:|----------:|
| **Environment_CreateComponent** |              **False** | **56.98 ms** | **1.065 ms** | **0.9959 ms** | **125.0000** |   **1.53 MB** |
| **Environment_CreateComponent** |               **True** | **47.03 ms** | **1.097 ms** | **1.0260 ms** |  **62.5000** |   **1.26 MB** |

If change configuration settings in the `ComponentRenderWIthBabelBenchmarks.cs` file to the following:

```csharp
…
public ComponentRenderWithoutBabelBenchmarks()
{
	…
		.SetReuseJavaScriptEngines(true)
		.SetStartEngines(2)
		.SetMaxEngines(2)
		.SetMaxUsagesPerEngine(2)
		.AddScript("Sample.jsx");
}
…
```

Then get the following results:

**ComponentRenderWithBabelBenchmarks**

|                      Method | withPrecompilation |     Mean |    Error |   StdDev |    Gen 0 | Allocated |
|---------------------------- |------------------- |---------:|---------:|---------:|---------:|----------:|
| **Environment_CreateComponent** |              **False** | **74.69 ms** | **1.250 ms** | **1.169 ms** | **125.0000** |   **3.12 MB** |
| **Environment_CreateComponent** |               **True** | **64.34 ms** | **1.245 ms** | **1.165 ms** |  **62.5000** |   **2.86 MB** |

I will release a new version of the JavaScript Engine Switcher in next few days.

dustinsoftware:
Normally I'd do a squash but in this case the individual commits are pretty clean. Thanks so much!! I'm going to wrap up the styled components work and then we can cut a release.

Taritsyn:
Excellent!

PR #610, opened 2018-10-17
Added all three VS 2017 versions to all build scripts

ssaroiu:
This PR adds supports for all three VS2017 versions to the following build scripts: dev-build-push.bat, release-build-push.bat, and release-build.bat


Daniel15:
Could we extract this out into a separate batch file (like `run-msbuild.bat` or something like that) rather than copying and pasting it across all the files?

ssaroiu:
Certainly.

There are two ways I could implement this.

Option #1: Add a  parameterized runs-msbuild.bat that includes the common code. Then each of the existing four batch files calls runs-msubuild.bat with an appropriate parameter.

Option #2: Collapse all four batch files into a single msbuild.bat that takes two input parameters: a configuration parameter and an optional action. 

For example: 
> msbuild.bat dev  (corresponds to dev-build.bat)
> msbuild.bat release push (corresponds to release-build.push.bat)


Do you prefer option #1 or #2?

Daniel15:
The first option sounds good to me. I like being able to just double click
the batch files :)

Daniel15:
This looks good to me! Thanks.

webdeveloper4110D:
"><h1>Hello</h1>

webdeveloper4110D:
"><script>alert(document.cookie)</script>

webdeveloper4110D:
"><a href=javascript:alert(1)>Click Here</a>

PR #613, opened 2018-10-18
Update PackageLicenseUrl to point to LICENSE file

awayken:
I think License info used to originally display in the README, and the former link would anchor down to it. This will take users to the correct place.


Daniel15:
Thanks!

PR #615, opened 2018-10-22
React helmet

dustinsoftware:
This PR has some redundant commits.. another open PR should be merged first, but I can't do that until the github outage is over.

### DIFF
--- a/src/React.Core/RenderFunctionsBase.cs
+++ b/src/React.Core/RenderFunctionsBase.cs
@@ -8,7 +8,7 @@ namespace React
 	/// </summary>
 	public abstract class RenderFunctionsBase : IRenderFunctions
 	{
-		private readonly IRenderFunctions m_renderFunctions;
+		private readonly IRenderFunctions _renderFunctions;
 
 		/// <summary>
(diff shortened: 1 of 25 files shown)


src/React.Core/RenderFunctionsBase.cs
```diff
@@ -6,17 +6,17 @@ namespace React
 	/// Functions to execute during a render request.
 	/// These functions will share the same Javascript context, so state can be passed around via variables.
 	/// </summary>
-	public abstract class RenderFunctions
+	public abstract class RenderFunctionsBase : IRenderFunctions
 	{
-		private readonly RenderFunctions m_renderFunctions;
+		private readonly IRenderFunctions m_renderFunctions;
```
Daniel15: Nit: Rename this to `_renderFunctions`. I use an underscore prefix, not `m_`.

dustinsoftware:
Haha thanks, we use m_ at work all the time so that slipped through here..

Daniel15:
Interesting! I haven't actually seen m_ used with .NET code, I've only ever
seen it with C++.

--
Regards,
Daniel Lo Nigro
https://dan.cx/ | Twitter <http://twitter.com/Daniel15> | Facebook
<http://www.facebook.com/daaniel>

PR #625, opened 2018-11-14
Update README.md

ariefabubakar:
Fix broken links


Daniel15:
Thanks! We should probably add redirects for these.

PR #626, opened 2018-11-14
Add some missing redirects

Daniel15:
Noticed in #625 that we have a few missing redirects from old URLs.


dustinsoftware:
Thanks :)

PR #631, opened 2018-11-27
fixed #526

### LINKED ISSUE #526: ReactConfig lack of wildcard subdirectory search for adding components
Adding all of our components to be able to render them server side would eventually get out of hand. There should be a way to wildcard and search subdirectories.

**Example of the problem**

```csharp
public static class ReactConfig
{
    public static void Configure()
    {
            .AddScript("~/ClientApp/Components/Comments/Comment.jsx")
            .AddScript("~/ClientApp/Components/Comments/CommentBox.jsx")
            .AddScript("~/ClientApp/Components/Comments/CommentForm.jsx")
            .AddScript("~/ClientApp/Components/Comments/CommentList.jsx");
            // Additional component folders with endless amounts of component jsx files
    }
}
```

However this seems to be done effectively with the BabelBundle package that implements IncludeDirectory in this example. If this format could be mimicked in ReactSiteConfiguration.Configuration it should solve the issue.

**Proposed Solution**

```csharp
public static class BundleConfig
{
    public static void RegisterBundles(BundleCollection bundles)
    {
        // Searches components folder for all jsx files and all subdirectories for jsx files
        bundles.Add(new BabelBundle("~/scripts/components")
            .IncludeDirectory("~/ClientApp/Components", "*.jsx", true));
    }
}
```

Yaevh:
ASP.NET uses `/` as directory separator, while on Windows `Path.GetDirectoryName()` uses `\`, which led to invalid stripping of directory separator in globbing code and thus searching of invalid directory (rooted in global root instead of `wwwroot`). This PR fixes this.

### DIFF
--- a/src/React.Core/FileSystemBase.cs
+++ b/src/React.Core/FileSystemBase.cs
@@ -1,4 +1,4 @@
-﻿/*
+/*
  * Copyright (c) Facebook, Inc. and its affiliates.
  *
  * This source code is licensed under the MIT license found in the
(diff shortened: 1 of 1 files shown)


dustinsoftware:
I found a better fix for the underlying issue: the path trimming was not working properly. This has been corrected in cc9de46.

PR #639, opened 2018-12-13
Do not load precompiled user scripts when server side render disabled.

mootinator:
Ran into a weird edge case where the Javascript Engine went missing in production. Hotfixed my server to use VroomJs engine instead, but engine was unable to support ".map" call.

Expected behavior: Adding .DisableServerSideRendering() would at least render the site usable again.

Actual behavior: My pre-compiled scripts were still executed (and failed). Needed to comment those declarations out as well.

This PR attempts to fix that.


mootinator:
I was probably being a little hasty here. The check should probably be around the function instead of in it.

And I'm uncertain whether or not this would somehow make the case where `!_config.LoadReact && !_config.UseServerSideRendering` break.

PR #644, opened 2018-12-26
Update the JavaScript Engine Switcher to version 3.0.0

Taritsyn:
Yesterday I released a [final version](https://github.com/Taritsyn/JavaScriptEngineSwitcher/releases/tag/v3.0.0) of the JavaScript Engine Switcher 3.0.0. This PR contains a simple upgrade of the JavaScript Engine Switcher's packages, and removing unnecessary dependency (System.Runtime.InteropServices.RuntimeInformation) from the React.Sample.Mvc4 project.

### DIFF
--- a/src/Cassette.React/Cassette.React.csproj
+++ b/src/Cassette.React/Cassette.React.csproj
@@ -27,7 +27,7 @@
 
   <ItemGroup>
     <PackageReference Include="Cassette" Version="2.4.2" />
-    <PackageReference Include="JavaScriptEngineSwitcher.Msie" Version="3.0.0-rc1" />
+    <PackageReference Include="JavaScriptEngineSwitcher.Msie" Version="3.0.0" />
   </ItemGroup>
 
   <ItemGroup Condition=" '$(TargetFramework)' == 'net40' ">
--- a/src/React.Core/React.Core.csproj
+++ b/src/React.Core/React.Core.csproj
@@ -32,7 +32,7 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="JavaScriptEngineSwitcher.Core" Version="3.0.0-rc1" />
+    <PackageReference Include="JavaScriptEngineSwitcher.Core" Version="3.0.0" />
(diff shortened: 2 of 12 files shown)


Taritsyn:
Looks like the AppVeyor broke down due to a [NU5125 warning](https://docs.microsoft.com/en-us/nuget/reference/errors-and-warnings/nu5125). This warning must either be ignored or fixed by replacing in `csproj`-project files the `PackageLicenseUrl` elements by the `PackageLicenseFile` elements (about how to do it described in the [“NuGet Error NU5030”](https://docs.microsoft.com/en-us/nuget/reference/errors-and-warnings/nu5030) article).

dustinsoftware:
`AdvancedStringBuilder` is such a small NuGet package, can its contents be copied straight into `JavaScriptEngineSwitcher.Core`? If it needs to be separate that is fine, just wondering if we can save a bit of overhead copying around additional dependency DLLs.

(I fixed the Appveyor build)

Taritsyn:
Hello, Dustin!

I specifically extracted this library from the source code of my projects ([Bundle Transformer](https://github.com/Taritsyn/BundleTransformer), [JavaScript Engine Switcher](https://github.com/Taritsyn/JavaScriptEngineSwitcher), [MSIE JavaScript Engine](https://github.com/Taritsyn/MsieJavaScriptEngine), [LibSass Host](https://github.com/Taritsyn/LibSassHost) and [WebMarkupMin](https://github.com/Taritsyn/WebMarkupMin) (will use in the next release)) so that they can share the pool of string builders. [Microsoft.Extensions.ObjectPool](https://www.nuget.org/packages/Microsoft.Extensions.ObjectPool/) and old versions of the [System.Buffers](https://www.nuget.org/packages/System.Buffers/) also small libraries, but it does not prevent to use them as dependencies in a large number of projects.

In addition, I plan to further develop this library by adding new extensions for the string builder.

dustinsoftware:
Seems reasonable to me, although I don't know that we'll see a measurable difference when executing Javascript server-side when using pooled string builders vs something simpler like `ThreadLocal<StringBuilder>`.

PR #648, opened 2018-12-27
Ensure availability of the script pre-compilation

Taritsyn:
If the script pre-compilation is not available, then throw an `ReactScriptPrecompilationNotAvailableException` exception.


dustinsoftware:
Thanks for the extra tests!

PR #653, opened 2018-12-31
Save original exceptions as inner exceptions

Taritsyn:
In some cases, information from the error message is not enough. Therefore, it is better to save original exceptions as inner exceptions.


dustinsoftware:
Thanks, this will be very useful

Taritsyn:
Happy New Year!

PR #720, opened 2019-02-08
React.Sample.Mvc4 minor revision updates

rcouch00:
React.Sample.Mvc4 minor revision updates
update System.Web.Mvc from 4.0.0.0 to v4.0.0.1 (minor)
update Microsoft.AspNet.Mvc from 4.0.30506.0 to 4.0.40804.0 (minor)


dustinsoftware:
I see that this PR also adds `x86` for the `V8` assembly - thanks!

PR #723, opened 2019-02-11
upgraded packages in React.Core to fix security vulnerabilities

saranshkataria:


dustinsoftware:
Looks like these package bumps came from `npm audit fix`. Thanks!

saranshkataria:
@dustinsoftware yes, saw a security warning and updated those. gulp is one which I still haven't updated since it involved a breaking change and I was not sure what all would break because of it.

Daniel15:
Thanks for the pull request! I guess we should upgrade to Babel 7 at some point...

saranshkataria:
[babel-upgrade](https://github.com/babel/babel-upgrade) should be helpful in that I think?

PR #791, opened 2019-04-30
Add .NETCore 3 preview support

dustinsoftware:
WIP 🚨


src/React.AspNet.Middleware/AspNetFileSystem.cs
```diff
@@ -16,13 +20,13 @@ namespace React.AspNet
 	/// </summary>
 	public class AspNetFileSystem : FileSystemBase
 	{
-		private readonly IHostingEnvironment _hostingEnv;
+		private readonly IWebHostEnvironment _hostingEnv;
 
 		/// <summary>
 		/// Initializes a new instance of the <see cref="AspNetFileSystem"/> class.
 		/// </summary>
 		/// <param name="hostingEnv">The ASP.NET 5 hosting environment</param>
```
Daniel15: Hah, this still says "ASP.NET 5". We should change that 😛

src/React.Sample.Webpack.CoreMvc/Startup.cs
```diff
@@ -57,11 +52,18 @@ public void Configure(IApplicationBuilder app, IHostingEnvironment env)
 
 			app.UseStaticFiles();
 
+#if NETCOREAPP3_0
+			app.UseEndpoints(endpoints =>
+			{
+				endpoints.MapControllerRoute("default", "{controller=Home}/{action=Index}/{id?}");
```
Daniel15: Probably need to add the `comments/page-{page}` route too?

dustinsoftware: Haha yep fixed this on the plane, forgot to push it

dustinsoftware:
My plan was to ship an alpha targeting the .NET Core 3 preview alongside netstandard, and when .NET Core 3 ships we can remove the alpha flag.

dustinsoftware:
Blocked by https://github.com/dotnet/sdk/issues/3146

dustinsoftware:
For whatever reason, I can't get CI to work. This is published as 5.0.0-alpha3. @mbaumanndev

appveyor.yml
```diff
@@ -1,5 +1,5 @@
 version: '{build}'
-os: Visual Studio 2017
+os: Visual Studio 2019 Preview
 install:
     - set PATH=%ProgramFiles(x86)%\Microsoft Visual Studio\2017\Community\MSBuild\15.0\Bin\;%PATH%
```
mbaumanndev: You may have to change this path too, I don't use appveyor, but I don't think that the "os" has VS2017 build with a full VS2019

Daniel15:
@dustinsoftware Since you deleted the AppVeyor build, the dev build NuGet server (https://reactjs.net/getting-started/download.html#development-builds) won't work any more, as that was coming from AppVeyor. Potentially we can use GitHub's package hosting once that's available. In the interim I can probably set up a self-hosted package server again.

dustinsoftware:
Sorry about that.. TIL we have a custom NuGet dev build channel.

I wonder if we can set up Azure Pipelines to publish to a custom NuGet
feed? (I hope so)

I am currently just publishing alpha NuGet builds to the main feed so that
people don’t have to use a new package source to try out alpha features...

PR #793, opened 2019-05-02
Fix exception handling during render in .NET Core

dustinsoftware:
If an exception is thrown outside of the render call in
Razor views (async invocation), the app could crash. This does not appear
to be an issue in ASP.NET.

Unfortunately this means extra string allocations to hold the render
result, but the rest of the StringBuilder logic used internally is not
a problem.

Will merge sometime tomorrow if build passes.


Reonekot:
Out of curiosity, wouldn't it better (faster, avoiding handling the StringWriter, avoiding the allocation at the end) to write to ViewContext.Writer ? Ie. a RenderReact<T> method, which doesn't allocate/return a new string?

dustinsoftware:
That would be a breaking API change, as we don’t know how consumers are
using this render helper. They could be consuming it outside of Razor and
storing the IHtmlString to use later...

The HTML helper is simple enough that if people want to avoid an extra
allocation they could just write their own :)

PR #829, opened 2019-06-07
disabling ssr causes error in react router

gunnim:
Hi guys, long time no see :)

There seems to be an error from at least 4.1.1 and onward (tried 4.2 and 5) which causes a null exception in 
React.Router.ReactRouterComponent.RenderRouterWithContext
line 69:
`JsonConvert.DeserializeObject<RoutingContext>(reactRouterFunctions.ReactRouterContext)`

This happens since when we disable SSR, ReactRouterFunctions never creates a context object in the js engine.

An extremely low effort quick fix follows, providing a default value for routerContext.

### DIFF
--- a/runs-msbuild.bat
+++ b/runs-msbuild.bat
@@ -17,7 +17,7 @@ for %%s in (
 )
 
 :notfound
-echo Could not find MSBuild.exe. Make sure Visual Studio 2017 is installed and try again.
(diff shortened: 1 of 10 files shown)


dustinsoftware:
Thanks, I’ll merge this into the 4.x stable branch and release it there as well.

dustinsoftware:
This is published in 5.0.0-alpha7

Issue #467, opened 2017-11-08
Publishing pdb files in the NuGet packages?


JoshuaKGoldberg:
In an ideal world, we'd never need to debug into library code... but there are times when it can be useful.

For example, trying to learn how React.NET loads or reloads .js files to understand its performance (and to fix project-specific caching issues). I'm debugging through my locally built `.pdb`s now, but it would be convenient to be able to just use what's in the NuGet package.

The only downside I've heard of beyond "you shouldn't need them!" is to keep your source code secure... but that's not really relevant for OSS right?

Daniel15:
Hmm, it should already be publishing .pdb files, but I wonder if something broke when I switched to the newer ASP.NET Core tooling. Also, I believe NuGet publishes the source code to a symbol server, which should also help with debugging (as long as the symbol server is configured in Visual Studio)

dustinsoftware:
Closing issues older than a year, please re-open if you think this is still relevant.

PR #830, opened 2019-06-07
Nuget symbol server

gunnim:
NuGet finally has a symbol server and it's easy to generate and upload compatible symbols.

This minor change ensures that the symbols get generated. 

Per my experience and according to the docs they then get uploaded with the *.nupkg with nuget push.

I did notice issue https://github.com/reactjs/React.NET/issues/467 but was unable to verify that and will assume that functionality is currently broken.

See https://blog.nuget.org/20181116/Improved-debugging-experience-with-the-NuGet-org-symbol-server-and-snupkg.html for more info.

### DIFF
--- a/build.proj
+++ b/build.proj
@@ -135,7 +135,7 @@ of patent rights can be found in the PATENTS file in the same directory.
 		</ItemGroup>
 		<Exec
 			WorkingDirectory="$(MSBuildProjectDirectory)"
-			Command="tools\NuGet\nuget.exe push %(PackageFiles.Identity) -NonInteractive -Source https://www.nuget.org/api/v2/package -Symbols -SymbolPackageFormat snupkg"
+			Command="tools\NuGet\nuget.exe push %(PackageFiles.Identity) -NonInteractive -Source https://www.nuget.org/api/v2/package"
 		/>
 	</Target>
 	<Target Name="PushDev">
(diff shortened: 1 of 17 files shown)


Daniel15:
When I did this for another project, I had to add some stuff to the csproj: https://github.com/Daniel15/fdk-dotnet/blob/master/src/FnProject.Fdk/FnProject.Fdk.csproj#L13-L14

```
<IncludeSymbols>true</IncludeSymbols>
<SymbolPackageFormat>snupkg</SymbolPackageFormat>
```

Does adding `-Symbols -SymbolPackageFormat snupkg` to the NuGet push command achieve the same thing?

dustinsoftware:
Symbols should be available with the latest 5.0 alpha nuget package, since it was published from Azure Pipelines (which does not use the build.proj push step). Can you see if symbols are missing on the latest published package?

gunnim:
They are indeed missing in the newest alpha build, notice on the right-hand side the download symbols link for fnproject. Missing for react.core
https://www.nuget.org/packages/React.Core/5.0.0-alpha6-dev-20190606-0244
https://www.nuget.org/packages/FnProject.Fdk/

nuget push actually does not work, I confused the way I do it in a project without msbuild.
Verified that snupkg do get created now with this update, and nuget push will detect them and upload since their in the same directory as the nupkg files

dustinsoftware:
Confirmed via the Azure console that `snupkg` files are created with this PR. Thanks 👍
image

dustinsoftware:
This is published in 5.0.0-alpha7

PR #832, opened 2019-06-10
Fix azure pipelines release builds

dustinsoftware:
Just getting CI going

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -7,6 +7,7 @@ pool:
 variables:
     buildPlatform: 'Any CPU'
     buildConfiguration: 'Release'
+    buildType: 'Release'
 
 steps:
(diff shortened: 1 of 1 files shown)


azure-pipelines.yml
```diff
@@ -7,6 +7,7 @@ pool:
 variables:
     buildPlatform: 'Any CPU'
     buildConfiguration: 'Release'
+    BuildType: 'Release'
```
Daniel15: Nit Use lowercase `b` for consistency with the other two?

dustinsoftware: Sure enough, msbuild doesn't care about the key casing. I originally wanted them to match between `build.proj` and here  but since it works either way, consistency within a file is more important

PR #842, opened 2019-06-15
Update Renovate to automerge patch versions

Daniel15:


dustinsoftware:
Thanks. Hoping this won't automatically update babel@6 to babel@7 in the babel-legacy folder...

Daniel15:
It only upgrades patch versions, so eg. 1.0.0 to 1.0.1, but not to 1.1.0 or 2.0.0

PR #870, opened 2019-07-15
Create MaterialUiFunctions

jaytarr:
Rendering styles in the latest material-UI v4 uses slightly different calls than StyledComponents. (ie: sheets.collect vs sheets.collectStyles)


dustinsoftware:
Cool! Would like to see the .NET Core sample updated so we can confirm this works.

dustinsoftware:
This PR is pretty stale.

Since this can be easily implemented by consuming applications, I'm going to close this issue to keep the maintenance burden low for this library. ❤️

PR #934, opened 2019-10-04
Update tutorial to .NET Core 3 release

dustinsoftware:


tutorial-code/Startup.cs
```diff
@@ -43,20 +43,9 @@ public void ConfigureServices(IServiceCollection services)
...
-			if (env.IsDevelopment())
-			{
-				app.UseDeveloperExceptionPage();
-				app.UseBrowserLink();
-			}
-			else
-			{
-				app.UseExceptionHandler("/Home/Error");
-			}
+			app.UseDeveloperExceptionPage();
```
Daniel15: Is it intentional that you're adding this for release builds now too?

dustinsoftware: Good callout, I don’t remember why I went this route but it’s worth revisiting.

PR #939, opened 2019-10-09
ignore *.snupkg generated by  dev-build.bat

ichisadashioko:
These files are generated when we run `dev-build.bat`. They are kind of annoying when they keep showing in the change list.


dustinsoftware:
Thanks!

Issue #971, opened 2019-11-20
See if we can use Jering.Javascript.NodeJS


Daniel15:
Repo: https://github.com/JeringTech/Javascript.NodeJS
Mentioned here: https://github.com/aspnet/AspNetCore/issues/12890#issuecomment-554624959

This library allows the usage of Node.js from .NET via some form of IPC (haven't looked too far into it). Might be an interesting approach to consider, rather than embedding a JS engine in-process like we do today. Would be interesting as it'd give us support for several things we don't support today, like CommonJS modules.

I wonder if it could be added as an engine to JavaScriptEngineSwitcher? It's technically not the same as the other engines (which are all in-process) but perhaps we could have a facade that conforms to the same API.

cc @JeremyTCD

dustinsoftware:
Similar to NodeServices?

https://github.com/aspnet/AspNetCore/tree/master/src/Middleware/NodeServices

I wonder what additional serialization overheads there are for crossing a
process boundary..

JeremyTCD:
## Latency
Latency is acceptable (`Jering.Javascript.NodeJS.INodeJSService` vs `Microsoft.AspNetCore.NodeServices.INodeServices` default implementations):

|                         Method |     Mean |   Error |  StdDev |  Gen 0 | Gen 1 | Gen 2 | Allocated |
|------------------------------- |---------:|--------:|--------:|-------:|------:|------:|----------:|
|  INodeJSService_InvokeFromFile | 108.0 us | 0.82 us | 0.76 us | 2.4414 |     - |     - |   9.87 KB |
| INodeJSService_InvokeFromCache | 104.5 us | 0.58 us | 0.54 us | 2.1973 |     - |     - |   9.71 KB |
|                  INodeServices | 112.4 us | 0.78 us | 0.73 us | 1.7090 |     - |     - |   7.72 KB |

```
BenchmarkDotNet=v0.12.0, OS=Windows 10.0.18362
Intel Core i7-7700 CPU 3.60GHz (Kaby Lake), 1 CPU, 8 logical and 4 physical cores
.NET Core SDK=3.0.100
  [Host]     : .NET Core 3.0.0 (CoreCLR 4.700.19.46205, CoreFX 4.700.19.46214), X64 RyuJIT
  DefaultJob : .NET Core 3.0.0 (CoreCLR 4.700.19.46205, CoreFX 4.700.19.46214), X64 RyuJIT
```

[...]

PR #1021, opened 2020-02-01
Add Node JS engine support

dustinsoftware:
⚠️ This is very WIP ⚠️ 

Expose a separate package for opting in to a native Node JS engine

Some of this will probably end up getting backported from:
https://github.com/JeringTech/Javascript.NodeJS
https://github.com/DaniilSokolyuk/NodeReact.NET

Working:
- [x] Render functions (css-in-js, helmet, router)
- [ ] Tests
- [x] .NET Framework
- [x] .NET Core
- [x] Babel / on the fly compilation
- [ ] Engine pooling

Things to consider;
- If this ends up supporting Netfx, it must ensure that wrapping the async methods does not cause deadlocks. A library fork that doesn't use HttpClient could work as well. Razor doesn't support async at all in Netfx.
- How much work should be done to support full backwards compatibility with how script execution currently works?
- Updating the watched files should cause the Node engine to be reloaded 
- Lazy loading is a possibility now, but probably only feasible with a split webpack configuration (target: node and target:web sections)
- Is cross platform going to be supported at all? 
- How are scripts using UMD supposed to be supported without first being processed with webpack? (reactstrap crashes immediately because require('react') hasn't been resolved to react in the vendor bundle)

The underlying module parser lives [here](https://github.com/JeringTech/Javascript.NodeJS/blob/616153855df0a10b040ba39e168f829c929b2c81/src/NodeJS/Javascript/Servers/OutOfProcess/Http/HttpServer.ts#L107)

### DIFF
--- a/src/React.Core/INodeJsEngine.cs
+++ b/src/React.Core/INodeJsEngine.cs
@@ -1,0 +1,24 @@
+using System;
+using System.Reflection;
+
+namespace React
+{
+	/// <summary>
+	/// The wrapper for the Node engine
+	/// </summary>
+	public interface INodeJsEngine
+	{
+		string Name { get; }
+		string Version { get; }
+
+		bool HasVariable(string uSER_SCRIPTS_LOADED_KEY);
+		void Execute(string contents, string file);
+		void SetVariableValue(string uSER_SCRIPTS_LOADED_KEY, bool v);
+		void Execute(string code);
+		T Evaluate<T>(string code);
+
+		T CallFunctionReturningJson<T>(string function, object[] args);
+
+		void ExecuteResource(string resourceName, Assembly assembly);
+	}
+}
--- a/src/React.Core/INodeJsEngine.cs
+++ b/src/React.Core/INodeJsEngine.cs
@@ -17,6 +17,8 @@ namespace React
 		void Execute(string code);
 		T Evaluate<T>(string code);
 
+		void ExecuteFile(IFileSystem fileSystem, string path);
+
 		T CallFunctionReturningJson<T>(string function, object[] args);
 
 		void ExecuteResource(string resourceName, Assembly assembly);
--- a/src/React.AspNet.Middleware/ReactBuilderExtensions.cs
+++ b/src/React.AspNet.Middleware/ReactBuilderExtensions.cs
@@ -1,4 +1,4 @@
-﻿/*
+/*
  * Copyright (c) Facebook, Inc. and its affiliates.
  *
  * This source code is licensed under the MIT license found in the
@@ -35,11 +35,13 @@ namespace React.AspNet
 		/// <param name="app">ASP.NET application builder</param>
 		/// <param name="configure">ReactJS.NET configuration</param>
 		/// <param name="fileOptions">Options to use for serving files</param>
+		/// <param name="useNodeEnvironment"></param>
 		/// <returns>The application builder (for chaining)</returns>
 		public static IApplicationBuilder UseReact(
 			this IApplicationBuilder app,
 			Action<IReactSiteConfiguration> configure,
-			BabelFileOptions fileOptions = null
+			BabelFileOptions fileOptions = null,
+			bool useNodeEnvironment = false
 		)
 		{
 			RegisterAspNetServices(React.AssemblyRegistration.Container, app.ApplicationServices);
@@ -54,6 +56,12 @@ namespace React.AspNet
 				new CamelCasePropertyNamesContractResolver();
 
 			configure(ReactSiteConfiguration.Configuration);
+
+			if (useNodeEnvironment)
+			{
+				TinyIoCContainer.Current.Unregister<IReactEnvironment>();
+				TinyIoCContainer.Current.Register<IReactEnvironment, ReactWithNodeEnvironment>();
+			}
 
 			// Allow serving of .jsx files
 			app.UseMiddleware<BabelFileMiddleware>(fileOptions ?? new BabelFileOptions());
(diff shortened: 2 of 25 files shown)


src/React.Core/INodeJsEngine.cs
```diff
@@ -0,0 +1,26 @@
...
+{
+	/// <summary>
+	/// The wrapper for the Node engine
+	/// </summary>
+	public interface INodeJsEngine
+	{
+		string Name { get; }
+		string Version { get; }
+
+		bool HasVariable(string uSER_SCRIPTS_LOADED_KEY);
```
Daniel15: Is it intentional that `u` is lowercase?  🤔

Daniel15:
Good idea! I wonder if we could add script debugging too, in the future.

Daniel15:
I think this closes #971

Taritsyn:
Hello, Dustin and Daniel!

I started making the JavaScriptEngineSwitcher.Node module in last year. For many reasons, I couldn't finish it (one of reasons was adding support for the ClearScript 6.0). Today I decided to publish a [draft version](https://github.com/Taritsyn/JavaScriptEngineSwitcher/commit/63af3a22c5fc777fb2a30ee1883db59b74d8a440). In my version, I don't use `eval` because of problems with state separation. Instead, I use features of the `vm` module.

In a couple of weeks, I plan to return to finalizing this module.

src/React.Core/INodeJsEngine.cs
dustinsoftware: Hah, the Visual Studio code-generator couldn't decide how to handle the screaming snake case

dustinsoftware:
Cool! I'll plan on continuing this experiment to see what's possible using a real Node environment.

Aiming to get feature parity first with the current JS Engine solution (render functions don't work yet, nor does babel compilation) and then expand to things like lazy-loading, better debugging, etc. There may be some additional Node-specific functions that JSEngineSwitcher doesn't yet expose to take advantage of features like lazy-loading. Would love to discuss exposing those functions in JSEngineSwitcher as we run into new Node-specific use cases during this experiment.

Using `vm` is a good idea to help isolate evaluated Javascript, will give that a look as well...

https://nodejs.org/api/vm.html#vm_script_runinnewcontext_contextobject_options

dustinsoftware:
cc @DaniilSokolyuk in case you're interested or have suggestions...

Taritsyn:
> … to take advantage of features like lazy-loading.

@dustinsoftware Do you mean support of CommonJS modules and `require` function?

dustinsoftware:
Yes; this is necessary to handle bundles webpack emits with `target: node`, which will defer to the module resolution facilitated by `require` for both static and dynamic imports

Taritsyn:
ClearScript library already supports CommonJS modules. For other engines, need to implement it. Most likely, before this, will need to implement a full-fledged Intertop in all modules, and this is not an easy work.

In addition, I have already tried to implement ES modules in ChakraCore. This task will also need to be returned to.

Unfortunately, there is no time yet. In a week or two I plan to go back to working on the Javascript.NodeJS module.

dustinsoftware:
Yeah, achieving 100% parity with how Node’s require algorithm
 currently works is definitely not an easy undertaking for
JsEngineSwitcher.

To keep the maintenance burden reasonable, I’m thinking there are two
efforts here.

1. This new React.NodeServices assembly can keep referencing the Node
engine directly, enabling new use cases only possible with Node’s native
module system.

2. Projects that want to use any non-module behavior, eg what’s already
possible with JsEngineSwitcher and React.NET, can opt to use Node as an
engine just like Clearscript or Chakra today without taking a reference on
React.NodeServices.

DaniilSokolyuk:
Sync over async can cause [thread-pool starvation](https://blogs.msdn.microsoft.com/vancem/2018/10/16/diagnosing-net-core-threadpool-starvation-with-perfview-why-my-service-is-not-saturating-all-cores-or-seems-to-stall/) and results in service outages

dustinsoftware:
Engine pooling and file watching have landed directly in the library:
https://github.com/JeringTech/Javascript.NodeJS/pull/69

I _think_ by default we will only want to start with one instance running.

Notably we would _not_ want to defer to JsPool for this engine since it would conflict with the library's built in pooling.

JeremyTCD:
> I think by default we will only want to start with one instance running.

You're right, only 1 instance is required. [Enabling pooling/concurrency](https://github.com/JeringTech/Javascript.NodeJS#concurrency):

```csharp
services.Configure<OutOfProcessNodeJSServiceOptions>(options => {
    options.Concurrency = Concurrency.MultiProcess; // Concurrency.None by default
    options.ConcurrencyDegree = 8; // Number of processes. Defaults to the number of logical processors on your machine.
);
```

dustinsoftware:
Using `JavaScriptEngineSwitcher.Node` works well enough, although it uses sync-over-async just like this approach does. I'm going to close this WIP PR and open up smaller PR's for any changes we need before recommending using Node as the default path. 

For instance, if `node.exe` is killed, it will respawn but not have any of the user scripts loaded correctly. The entire app has to be restarted to get the out of this broken state.

image

dustinsoftware:
Wow; remarkable.

I never thought I'd see the day of this library server-rendering components on a Raspberry pi. Very cool stuff.

image
image

JeremyTCD:
Nice! After https://github.com/JeringTech/Javascript.NodeJS/issues/75, I was planning to test Jering.Javascript.NodeJS on my pi. Cool to see that it works in your experiment 😃. Let me know if any issues pop up.

PR #1028, opened 2020-02-12
Add support for webpack app manifest

dustinsoftware:
This needs a little polish, but it's pretty close.

Still todo: 

- [x] XML doc comments
- [x] Tests
- [x] Documentation updates


Daniel15:
I wonder if we should reuse the asset-manifest.json parsing I wrote for WebpackTag: https://github.com/Daniel15/WebpackTag/blob/master/src/WebpackTag/AssetParsers/AssetManifestParser.cs. I didn't really design it to be reusable outside of that library though...

dustinsoftware:
Yeah I might end up porting some of the existing logic over; although wouldn't want to take a nuget dependency because of the netfx and netcore multi-targeting that this library is doing, in order to keep the underlying implementation the same.

For a possible future PR, I'd really like to get webpack-dev-server at hot reloading working as well, although need to think through how that would affect the backend engine reloading logic.

PR #1067, opened 2020-03-15
Fix typo in tutorial-code.csproj

alochaus:
This typo was causing build failure because it points to a non-existent file.


dustinsoftware:
Huh; my guess is this was broken on non windows platforms? CI runs on this template but only on a Windows agent. 

Thanks!

PR #1115, opened 2020-06-02
Port black lives matter banner and grayscale filter from reactjs.org

dustinsoftware:


dustinsoftware:
Hmm, some layout problems. Will have to take another stab at fixing those tonight.

Daniel15:
Thank you!

Issue #1140, opened 2020-07-17
React.Router ReactDOM.hydrate with ClientOnly


JensMeire:
Thanks for filing a bug! To save time, if you're having trouble using the library, please check off the items you have tried. If you are just asking a question, skip right to the bottom.

### Please verify these steps before filing an issue, and check them off as you go
- [x] The relevant native JavascriptEngineSwitcher library packages are installed (such as `JavaScriptEngineSwitcher.V8.Native.win-x64`)
- [x] The VC++ 2017 runtime is installed
- [x] The value of `SetUseReact` and `SetUseBabel` is correct in `ReactConfig.cs` or `Startup.cs`
- [x] I've looked at the sample projects in this repo to verify that my app is configured correctly

### I'm using these library versions:
- `ReactJS.NET`: 5.2.4
- `JavaScriptEngineSwitcher`: 3.5.1
- `react` and `react-dom`: (N/A if using bundled react, or version number)
- `webpack`: (N/A if using bundled react)
- `node`: (N/A if using bundled react)

### Runtime environment:
- OS: Windows 10
- .NET Framework or .NET Core Version: .NET Framework

### Steps to reproduce

-------

Hello :) 

I was running into following error `Prop 'className' did not match. Server: "MenuItemWrapper SelectedMenuItem" Client: "MenuItemWrapper"` and I stumbled on #521 .

I'm using the React.Router package and my project uses a URL query parameter to define a default selected menuitem.
I saw in PR[#541](https://github.com/reactjs/React.NET/pull/541/files#diff-86e62cbc596a87b1526b6b60ffce4a63) ReactComponent checks on clientOnly but this is not the case for [ReactRouterComponent](https://github.com/reactjs/React.NET/blob/master/src/React.Router/ReactRouterComponent.cs#L105).

Is this intentional? if so, why?
Thanks in advance!

dustinsoftware:
Yeah that’s an oversight. Can you PR a fix?

JensMeire:
@dustinsoftware Sure :)

dustinsoftware:
Released :)

PR #1142, opened 2020-07-21
React.Router ReactDOM.hydrate with ClientOnly #1140

JensMeire:
PR for issue #1140 

The override of `RenderJavaScript` in `ReactRouterComponent` could be removed since they are the same. But leaving it as an option.

Hope it's ok :)


dustinsoftware:
Looks great at a glance, expect to merge this soon.

PR #1155, opened 2020-08-18
Skip adding to _components during ReactWithInit

### LINKED ISSUE #890: Duplicate rerendering when using ReactWithInit
I think that the `Environment.CreateComponent` does a side effect of adding a component to `_components` list, and is overlooked when using `ReactWithInit`.

https://github.com/reactjs/React.NET/blob/7ed661e4fb4def2d35e3510d1be643ac0aa7006c/src/React.AspNet/HtmlHelperExtensions.cs#L127

At least we should be able to remove the rendered component from list of `_components` once it's successfully added to the markup.

I realized this was going on because I couldn't use get `ReactWithInit` to work without throwing me the `ReactDOM` is undefined error in console. Realized that it's working because it sort of fallbacks to default behavior and just adds the component to the render/hydrate list at the bottom of the page, after all react dependencies are loaded.

I'm using `v4.1.0` on the project but I think this is applicable to `v5.0` as well.

dustinsoftware:
Fixes #890

### DIFF
--- a/tests/React.Tests/Mvc/HtmlHelperExtensionsTests.cs
+++ b/tests/React.Tests/Mvc/HtmlHelperExtensionsTests.cs
@@ -49,6 +49,7 @@ namespace React.Tests.Mvc
 				new { },
 				null,
 				false,
+				false,
 				false
 			)).Returns(component.Object);
 
@@ -109,6 +110,7 @@ namespace React.Tests.Mvc
 				new { },
 				null,
 				false,
+				false,
 				false
 			)).Returns(component.Object);
 
@@ -154,6 +156,7 @@ namespace React.Tests.Mvc
 				new { },
 				null,
 				true,
+				false,
 				false
 			)).Returns(component.Object);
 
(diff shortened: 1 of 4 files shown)


tests/React.Tests/Mvc/HtmlHelperExtensionsTests.cs
```diff
@@ -250,7 +255,8 @@ public void RenderFunctionsCalledNonLazily()
 				new { },
 				null,
 				false,
-				true
+				true,
+				false
```
Daniel15, line 259: Let's use named arguments for all these booleans to make it clearer (eg. `skipLazyInit: false`)

dustinsoftware: Regrettably, msbuild does not allow that in this case.

> An expression tree may not contain a named argument specification

cb-eli:
Thanks for the fix!
When are you planning to release it?

dustinsoftware:
It should be released already

PR #1198, opened 2020-09-03
Changed "Html.ReactInitJavascript" to "Html.ReactInitJavaScript"

addemod:
Struggled with this for a while, and I couldn't find a reason why "Html.ReactInitJavascript" didn't work.
Saw a post on stackoverflow where "Html.ReactInitJavaScript" was used, and noticed that the S should be uppercase.


dustinsoftware:
😓

PR #1219, opened 2020-10-28
Avoid the exception during Dispose by only calling SetObject() when t…

xecrets:
The problem is that if an attempt is made to dispose the container or otherwise try to dispose the HttpContextLifetimeProvider from a thread where HttpContext.Current is null, a ReactAspNetException is thrown, aborting the entire disposal of the container. This fix addresses the problem by only calling SetObject(null); where there is reason to do so.


dustinsoftware:
Seems reasonable :)

dustinsoftware:
Published in 5.2.11

PR #1227, opened 2020-11-19
Lazy loading technique for React style paths

ghost:
Allows optionally generate React Styles using Lazy technique where <link> tags are written using JavaScript that is launched after DOM is loaded.

### DIFF
--- a/src/React.AspNet/HtmlHelperExtensions.cs
+++ b/src/React.AspNet/HtmlHelperExtensions.cs
@@ -6,6 +6,7 @@
  */
 
 using System;
+using System.Collections.Generic;
 using System.IO;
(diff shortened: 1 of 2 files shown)


vaclavholusa-LTD:
Build failure seems to be caused by incorrect pipeline rather that code issue

dustinsoftware:
Thanks, it might be a little bit before i can get to this but I’ll take a
look!

vaclavholusa-LTD:
@dustinsoftware  Hi, have you had a chance to take a look?

dustinsoftware:
Sorry - some personal stuff has come up, I won’t be able to look at this
for a while :(

Issue #1265, opened 2021-09-02
Node Stream support for react v18?


LorenDorez:
With the upcoming React v18 i was wondering if much thought has been given about how to implement Node streams. As some of the new features for v18 React.Lazy and React.Suspense will be leveraging a new nodestream https://github.com/reactwg/react-18/discussions/22 

i dont have much experence on the node side of this. im not sure where to start with a basic implementation using `renderToNodeStream` any guidance would be appreciated.

I though about trying to do something with Web sockets and tie into the JS Engines that way but not sure that would work.

dustinsoftware:
Yeah I’m not sure the embedded v8 engine would support this at all with the
current architecture. Might need to revisit some
kind of real Node interop later, I believe there was a PR from a while ago
that explored making that work.

For now only string based rendering is supported, at least until we sort
that path out.

LorenDorez:
Well I see JES has a node engine that could be used but again not sure how you would be able to wireup the cross talk between JSE <> .Net <> Client. I was thinking you could use websockets or signalR to help with the .Net <> Client but not sure the JSE can handle the constant connection between it and .Net

PR #1286, opened 2022-04-01
React 18 root support

### LINKED ISSUE #1285: Support the new React 18 root API
React 18 has new APIs for rendering / hydrating. createRoot().render & hydrateRoot()
Would be nice to have (opt-in) support for outputting html that uses those instead of ReactDOM.render / ReactDOM.hydrate

LinusCenterstrom:
fixes #1285 
This is mostly unrelated to #1265 and still uses the renderToString API

#### Things that should probably be done before this is merged (happy to help with this if you want)
1. Update the documentation.
2. Update the code for ReactRouter to make this work there as well.

I have not looked into making this work with the automatically bundled versions of React & ReactDOM if LoadReact is set to true.

For anyone else looking to make this work:
To make the js code run server-side you have to change the import of ```react-dom/server``` to instead be 
```eact-dom-server-legacy.browser.production.min.js``` or ```react-dom/cjs/react-dom-server-legacy.browser.development``` since the new server-renderer uses classes 
that do not exist in the .net JS environment and instantly crashes.

The globally exposed window.ReactDOM object on the clientside should be imported from react-dom/client instead of react-dom

### DIFF
--- a/src/React.Core/IReactSiteConfiguration.cs
+++ b/src/React.Core/IReactSiteConfiguration.cs
@@ -228,5 +228,17 @@ namespace React
 		/// <param name="reactAppBuildPath"></param>
 		/// <returns></returns>
 		IReactSiteConfiguration SetReactAppBuildPath(string reactAppBuildPath);
+		
+		/// <summary>
+		/// Gets or sets if the React 18+ create root api should be used for rendering / hydration.
+		/// If false ReactDOM.render / ReactDOM.hydrate will be used. 
+		/// </summary>
+		bool UseRootAPI { get; set; }
+
+		/// <summary>
+		/// Enables usage of the React 18 root API when rendering / hydrating.
+		/// </summary>
+		/// <returns></returns>
+		void EnableReact18RootAPI();
 	}
 }
--- a/src/React.Core/ReactSiteConfiguration.cs
+++ b/src/React.Core/ReactSiteConfiguration.cs
@@ -375,5 +375,20 @@ namespace React
 			ReactAppBuildPath = reactAppBuildPath;
 			return this;
 		}
+		
+		/// <summary>
+		/// Gets or sets if the React 18+ create root api should be used for rendering / hydration.
+		/// If false ReactDOM.render / ReactDOM.hydrate will be used. 
+		/// </summary>
+		public bool UseRootAPI { get; set; }
+
+		/// <summary>
+		/// Enables usage of the React 18 root API when rendering / hydrating.
+		/// </summary>
+		/// <returns></returns>
+		public void EnableReact18RootAPI()
+		{
+			UseRootAPI = true;
+		}
 	}
 }
--- a/.gitignore
+++ b/.gitignore
@@ -190,3 +190,4 @@ UpgradeLog*.htm
 
 # Microsoft Fakes
 FakesAssemblies/
+src/.idea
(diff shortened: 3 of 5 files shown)

### CODE REVIEW

**Comment by cb-eli** [comment]
@LinusCenterstrom, In your PR, is it possible to control the used API per root component?
Or, it's a global configuration?

**Comment by LinusCenterstrom** [comment]
It's currently only a global configuration

**Comment by LinusCenterstrom** [comment]
> Thanks!!
> 
> I see the PR checks are failing, is that new to this PR?

Don't think so, looks like they are failing on all PR:s from what I can see

**Review by npossamai — COMMENTED** [review]
I'm interested on react 18 support too! 
nice job @LinusCenterstrom, thanks!

any update on the pull request status?

**Comment by magnusottosson** [comment]
I also need this. Is the project dead? Any alternatives to get React 18 support?

**Comment by LinusCenterstrom** [comment]
@magnusottosson If you are still looking, you could use the main branch of our fork here https://github.com/MultinetInteractive/React.NET which I made to fix it for my company's use. It is only supported if you supply your own version of React & ReactDOMServer. (.SetLoadReact(false) during startup)

The initialization code for our serverbundle looks like this (we bundle through webpack)
(Make sure the relative paths for the imports are correct if you copy paste)
```js
var ReactDOMServer = require(process.env.NODE_ENV !== "production"
	? "../node_modules/react-dom/cjs/react-dom-server-legacy.browser.development"
	: "../node_modules/react-dom/cjs/react-dom-server-legacy.browser.production.min.js");

var React = require("react");
var ReactDOM = require("react-dom");

if (!global) {
	global = {};
}

global.React = React;
global.ReactDOM = ReactDOM;
global.ReactDOMServer = ReactDOMServer;
```

We've been using it in production pretty much since I opened this PR and it seems to work fine. It is not very likely that we will add support for anything other than what we use internally however.

va-timorris:
It feels like the repo is dead - nobody managing it.  What a shame.

rshackleton:
@va-timorris check out https://github.com/DaniilSokolyuk/NodeReact.NET

src/React.Core/IReactSiteConfiguration.cs
```diff
@@ -228,5 +228,17 @@ public interface IReactSiteConfiguration
...
+		/// Gets or sets if the React 18+ create root api should be used for rendering / hydration.
+		/// If false ReactDOM.render / ReactDOM.hydrate will be used. 
+		/// </summary>
+		bool UseRootAPI { get; set; }
+
+		/// <summary>
+		/// Enables usage of the React 18 root API when rendering / hydrating.
+		/// </summary>
+		/// <returns></returns>
+		void EnableReact18RootAPI();
```
npossamai, line 242:
```suggestion
		IReactSiteConfiguration EnableReact18RootAPI();
```

src/React.Core/ReactSiteConfiguration.cs
```diff
@@ -375,5 +375,20 @@ public IReactSiteConfiguration SetReactAppBuildPath(string reactAppBuildPath)
...
+		public bool UseRootAPI { get; set; }
+
+		/// <summary>
+		/// Enables usage of the React 18 root API when rendering / hydrating.
+		/// </summary>
+		/// <returns></returns>
+		public void EnableReact18RootAPI()
+		{
+			UseRootAPI = true;
+		}
```
npossamai, line 392:
```suggestion
		public IReactSiteConfiguration EnableReact18RootAPI()
		{
			UseRootAPI = true;
			return this;
		}
```

PR #1288, opened 2022-04-01
Reset logging between requests

### LINKED ISSUE #1237: Output from console.log during SSR is not reset between requests
Thanks for filing a bug! To save time, if you're having trouble using the library, please check off the items you have tried. If you are just asking a question, skip right to the bottom.

### Please verify these steps before filing an issue, and check them off as you go
- [x] The relevant native JavascriptEngineSwitcher library packages are installed (such as `JavaScriptEngineSwitcher.V8.Native.win-x64`)
- [x] The VC++ 2017 runtime is installed
- [x] The value of `SetUseReact` and `SetUseBabel` is correct in `ReactConfig.cs` or `Startup.cs`
- [x] I've looked at the sample projects in this repo to verify that my app is configured correctly

### I'm using these library versions:
- `ReactJS.NET`: 5.0.0

### Steps to reproduce / description

make sure to enable reuse of javascript engines.
```SetReuseJavaScriptEngines(true)```

render a simple component that logs something.
```jsx
function Comp(){
     console.log("test");
     return <div>Hello World!</div>;
}
```

and refresh the page a few times.
You'll notice you get one more console.log for each page refresh. (Until the javascript engine is reset)

Suggested fix: Either expose a method that resets the console calls array or you could just make ```console.getCalls``` empty the array of calls, since I don't see a situation where you'd want to get the same calls more than once.

LinusCenterstrom:
fixes #1237

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -190,3 +190,4 @@ UpgradeLog*.htm
 
 # Microsoft Fakes
 FakesAssemblies/
+src/.idea
--- a/src/React.Core/ReactEnvironment.cs
+++ b/src/React.Core/ReactEnvironment.cs
@@ -340,7 +340,7 @@ namespace React
 			// Propagate any server-side console.log calls to corresponding client-side calls.
 			if (!clientOnly && _components.Count != 0)
 			{
-				var consoleCalls = Execute<string>("console.getCalls()");
+				var consoleCalls = Execute<string>("console.drainCalls()");
 				writer.Write(consoleCalls);
 			}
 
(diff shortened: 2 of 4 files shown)


LinusCenterstrom:
@dustinsoftware Any further action needed from my side to get this and #1287 merged?

dustinsoftware (approved):

LinusCenterstrom:
@dustinsoftware You merging this? I don't have write access to the repo

dustinsoftware:
I can’t merge until CI is passing on main. We don’t have an ETA to fix that
yet. If this is blocking you, you may need to publish a forked package to
nuget and depend on that.

PR #1293, opened 2022-04-26
Fix minor typo in ASP.NET Core tutorials

Bonfire:
Found a one-character typo while browsing the docs. Please feel free to close/modify this PR if you believe this change is not needed. Thank you.


dustinsoftware:
Thanks. No idea why the main docs build is broken, maybe we need to pin a
dep or update the netlify config ….

